PR #12, opened 2017-08-08
Add agenda item about IDs

stubailo:
I'm interested in resurrecting the ID proposal: https://github.com/facebook/graphql/pull/232

There's just so much that can be done if tooling can rely on a standard ID concept, I think it might be worthwhile to take a first step there.


wincent:
Thanks @stubailo.

PR #15, opened 2017-08-08
Add GCL to agenda

schickling:


wincent:
Thanks @schickling. Do you have a link that we could add to this item for context. I am not actually sure what you're talking about here; is it [graphql-config](https://github.com/graphcool/graphql-config)?

schickling:
I'll write something up before the meeting including a little demo. I'll create another PR! 🙂

PR #25, opened 2017-10-16
Adding myself to attendees

IvanGoncharov:
Not sure about the location though.
@wincent @leebyron Are you still planning to host the meetup the next day after GraphQL Summit?


leebyron:
We're going to host, yes! I'll update with location info

PR #27, opened 2017-10-19
Add Oleg Ilyenko

OlegIlyenko:
Excited about the next meeting! Unfortunately, I will not manage to attend the GraphQL Summit this time around. Would it still be possible to join the meeting remotely?


leebyron:
Yes! We will have a remote dial in :)

PR #29, opened 2017-10-20
add andi marek

andimarek:
It is in the middle of the night and I will try to join. :-)


leebyron:
It's tough to get one time that works for all, but it would be great to have you

PR #30, opened 2017-10-21
Add Apollo HQ location and attendees

martijnwalraven:


leebyron:
Thanks!

You're all more then welcome to join us at the FB location as well. Since we have a bunch of people coming in for Summit, we booked a large room and are taking people to lunch afterwards.

PR #32, opened 2017-10-23
Add a few proposed topics to the agenda

martijnwalraven:


leebyron:
Great topics, thanks for filling out the agenda!

PR #34, opened 2017-10-26
Add "GraphQL over HTTP" specification to agenda

IvanGoncharov:
GraphQL spec doesn't document transport layer and I think it's great. It allows a lot of flexibility and experimentation.

It's great that we have a standard for query and responses but you should be able to send those in a standardized manner. We faced this issue while working on a few GraphQL proxy tools at APIs.guru and we think it may be a limiting factor for the growth of GraphQL tooling.

My proposal is to create separate "GraphQL over HTTP" specification based on [express-graphql](https://github.com/graphql/express-graphql) and just document existing practices as the first step.


wincent:
Thanks, @IvanGoncharov.

PR #38, opened 2017-11-13
please add glassdoor to the agenda

sid88in:


sid88in:
We are leading contributors to serverless graphql - https://github.com/serverless/serverless-graphql-apollo 

@schickling invited us to be a part of this. Would be great to share thoughts and work together!

sid88in:
would be great if you could add me to the group!

thanks

leebyron:
The agenda you're editing is from a past meeting! If you'd like to join the next please add your name to the correct date.

PR #43, opened 2018-01-28
WG Meeting #3 attendee: Mike Marcacci

mike-marcacci:
Adding myself [as suggested](https://github.com/facebook/graphql/issues/295#issuecomment-347714235) to talk about allowing interfaces to implement other interfaces.


leebyron:
Great! Please PR to add that topic to the agenda items so we don't forget to cover it

PR #45, opened 2018-01-29
Add discussion points to the agenda

xuorig:
Adding two discussion points to the agenda.

They're both quite large subjects so I put 30m but feel free to scale that down to leave room for other subjects if needed 🙏 

cc @leebyron

### DIFF
--- a/agendas/2018-02-01.md
+++ b/agendas/2018-02-01.md
@@ -44,6 +44,8 @@ Mike Marcacci        | Boltline      | San Diego, CA
 1. Determine volunteers for note taking (5m)
 1. Discuss upcoming meeting schedule (5m)
 1. Present [graphql-import](https://github.com/graphcool/graphql-import) (5m)
+1. Discuss exposing schema metadata (30m)
+1. Discuss errors (30m)
 
 ## Agenda and Attendee guidelines
 


agendas/2018-02-01.md
```diff
@@ -44,6 +44,8 @@ Mike Marcacci        | Boltline      | San Diego, CA
 1. Determine volunteers for note taking (5m)
 1. Discuss upcoming meeting schedule (5m)
 1. Present [graphql-import](https://github.com/graphcool/graphql-import) (5m)
+1. Discuss exposing schema metadata (30m)
+1. Discuss errors (30m)
```
xuorig: I know this has been discussed previously, feel free to remove it if you feel like this won't add much to the conversation 👍

leebyron:
Thanks!

agendas/2018-02-01.md
```diff
@@ -44,6 +44,8 @@ Mike Marcacci        | Boltline      | San Diego, CA
 1. Determine volunteers for note taking (5m)
 1. Discuss upcoming meeting schedule (5m)
 1. Present [graphql-import](https://github.com/graphcool/graphql-import) (5m)
+1. Discuss exposing schema metadata (30m)
```
OlegIlyenko: Great, thanks a lot for adding this agenda topic, @xuorig! 👍 I was also thinking about adding it, should be an interesting discussion.

IvanGoncharov:
@xuorig It would be great to have a short description of both agenda items. I think it allows attendees to think about them beforehand so we will have a more productive discussion.
Having agenda items link to GitHub comments similar to [previous agenda](https://github.com/graphql/graphql-wg/blob/master/agendas/2017-10-27.md#agenda) would be great.

> Discuss exposing schema metadata 

Do you mean something similar to https://github.com/facebook/graphql/issues/300 ?
Or you proposing to allow fully customizable introspection 
 as described here: https://github.com/rmosolgo/graphql-ruby/blob/1.8-dev/guides/schema/class_based_api.md#custom-introspection

xuorig:
@IvanGoncharov sure, good idea. I can add a bit of context here:

I'd like to be clear that I am adding these to the agenda to discuss potential solutions, not to propose one, so both links could be related. Issue #300 does talk a lot about the problem and is probably a good read before the work group meeting. Amongst other things, I'd like to discuss:

  - Is the graphql schema the right place for exposing metadata? Should it live outside the schema?
  - extending the introspection schema & custom Internal types `__MyCustomMetadata`
  - IDL vs JSON introspection
  - "global" metadata vs member metadata
  - directive limitations

For the errors point, I'd like to discuss a few points:

  - Using "client errors" (Defined in the schema) for mutations vs the spec's "errors" key.
  - Limitations to common client errors implementations.
  - Potential additions to the errors spec

(I can also talk about our specific use cases at the meeting.)

PR #51, opened 2018-01-30
Add myself, and add that I can take notes

orta:
Re: #50 

I have opinions on a bunch of this stuff, but nothing too strong, so I think taking notes would be a fine option for contributing. 

(Didn't add myself at the bottom to avoid merge conflicts)


leebyron:
> (Didn't add myself at the bottom to avoid merge conflicts)

Thanks for making my life easier :)

PR #52, opened 2018-01-30
Add Alan to attendees and live queries to agenda

acjay:
I'm interested in attending in-person. But I assumed location to be where I work. Not sure!

Also, didn't want to block too much time for discussion of live queries. I think 15 mins should be sufficient to share the findings of initial research, the next steps, and get some feedback.

I'm happy to take notes on another agenda section.


leebyron:
Great thanks! We'll have a video chat to dial into

acjay:
@leebyron I'm actually interested in attending in person, if possible. I happen to be in SF this week.

PR #59, opened 2018-02-01
Adds meeting notes

orta:


orta:
These are in 👍

notes/2018-02-01.md
```diff
@@ -0,0 +1,220 @@
...
+  - -ID field make sense?
+  - -Does the Relay ID make sense?
+  - -Does pagination make sense?
+
+  - In favour of splitting into more multiple sections:
+
+   Cursor per item feels weird to be normative
+   Leeb has a doc that is basically this, which has other alternatives to connections
+   C: Perhaps “pattern” as a term, is a best option?
+   Relay modern doesn’t use the global ID
```
alloy: This isn’t accurate. More accurate is that it uses an `id` field if it exists, otherwise generates its own IDs

For single nodes https://github.com/facebook/relay/blob/104b2c3a1e5199d7b2ba5c6e4ce0adffa6461fad/packages/relay-runtime/store/RelayResponseNormalizer.js#L259 and for plural nodes https://github.com/facebook/relay/blob/104b2c3a1e5199d7b2ba5c6e4ce0adffa6461fad/packages/relay-runtime/store/RelayResponseNormalizer.js#L308 (edited)

PR #69, opened 2018-03-22
Won't be able to attend :(

xuorig:
assumed wrongly it was at the same time as last time. I'll be there a the next one!

cc @leebyron


leebyron:
No problem! See you on the next one

PR #70, opened 2018-03-22
Update 2018-03-22.md

sorenbs:
- Present OpenCRUD
- status update on Top level directives (we need a champion)


leebyron:
Thanks!

PR #79, opened 2018-05-08
Time zone DST adjustments for WG #5

OlegIlyenko:
I would like to adjust the time based on the DST time zones since PDT is currently in effect.

time_converter_and_world_clock_-_conversion_at_a_glance_-_pick_best_time_to_schedule_conference_calls__webinars__online_meetings_and_phone_calls_


leebyron:
Thank you for catching and fixing this! And thanks for including UTC as a standard point to compare from.

PR #87, opened 2018-06-01
Update month for GQL Europe meeting from July to June

dwwoelfel:


IvanGoncharov:
@dwwoelfel Thanks 👍 
Same as #86 
I will merge this one before we get the third PR 😄 
I don't think it has to be reviewed.

PR #93, opened 2018-06-13
WG meeting #6 attendee: Matt Mahoney

mjmahone:
On the Location, I'm not sure whether that means where I'll be for the meeting, where I *usually* am, or where the company is based out of. Putting where I am *currently* rather than my "home base".

I could also reasonably host 1-3 people at Facebook for this WG meeting, if there was any interest in that. Probably couldn't easily host 10 people, though.


IvanGoncharov:
> On the Location, I'm not sure whether that means where I'll be for the meeting, where I usually am, or where the company is based out of. 

AFAIK, it's a place where you will be so others can join you on site.

PR #115, opened 2018-10-19
Add self to attendees

mjmahone:
Will be attending in-person.

### DIFF
--- a/agendas/2018-11-08.md
+++ b/agendas/2018-11-08.md
@@ -19,6 +19,7 @@ Name                 | Organization  | Location
 -------------------- | ------------- | ----------------------
 Lee Byron            | Robinhood     | Menlo Park, CA
 Michael Staib*       | ChilliCream   | Zurich, Switzerland
+Matt Mahoney         | Facebook      | Menlo Park, CA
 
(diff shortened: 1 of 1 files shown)


hwillson:
@mjmahone @leebyron Out of curiosity, do you have a physical location for the meeting set? We'd be happy to host the meeting at Apollo HQ if you don't yet.

cc @martijnwalraven @jbaxleyiii

leebyron:
I don't have a physical location - I'll personally not be able to make it to a physical meetup, but I think having space for one for those who can make it would be valuable!

IvanGoncharov:
> We'd be happy to host the meeting at Apollo HQ if you don't yet.

@hwillson It would be great 👍I will definetly join you for the call.
You can add location to the agenda similar to what we had for GraphQL Europe:
https://github.com/graphql/graphql-wg/blob/master/agendas/2018-06-14.md#physical-locations

abernix:
@IvanGoncharov 	@hwillson 	@leebyron 	I've opened a PR which suggests Apollo HQ for the physical location: https://github.com/graphql/graphql-wg/pull/121 We're happy to host all attendees!

PR #121, opened 2018-11-05
Add physical location for Apollo HQ

abernix:
Apollo is happy to host the GraphQL WG and host any of its attendees who might be in San Francisco.  We have a large conference room with video conferencing equipment reserved that can be used by remote attendees.

Ref: https://github.com/graphql/graphql-wg/pull/115#issuecomment-435749665

cc @leebyron @IvanGoncharov


leebyron:
Great!

abernix:
(We'll PR the video conference link closer to the event.)

PR #139, opened 2019-04-25
Adds myself to the May WG

orta:
I figure it can be a launching point to get some names and a date for scheduling a chat just about graphiql

### DIFF
--- a/agendas/2019-05-02.md
+++ b/agendas/2019-05-02.md
@@ -30,6 +30,7 @@ to the list of attendees below along with your organization and location. Add an
 Name                 | Organization       | Location
 -------------------- | ------------------ | ----------------------
 Lee Byron            | GraphQL Foundation | Menlo Park, CA
+Orta Therox          | Individual Contrib | New York, NY
 *ADD YOUR NAME HERE* | *COMPANY / ORG*    | *WHERE YOU ARE*
 
 <small>\*: willing to take notes (eg. Joe Montana\*)</small>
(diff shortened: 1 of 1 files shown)


agendas/2019-05-02.md
```diff
@@ -41,4 +42,5 @@ Lee Byron            | GraphQL Foundation | Menlo Park, CA
 1. Introduction of attendees
 1. Review agenda (5m)
 1. Determine volunteers for note taking (5m)
+1. Starting up the GraphiQL Resdesign WG (5-10m)
```
IvanGoncharov: @orta I would suggest generalizing this issue.
I think the core problem here is the lack of active maintainers on GraphiQL repo.
Even if we all agree with the proposed redesign, it doesn't solve the core issue of unmerged PRs and unanswered issues.

I would suggest making it 20m and focus more on how to bootstrap the long term "GraphiQL WG". Also, I think a redesign is not the best first project to form a group of dedicated maintainers. I would suggest just stabilizing repo and updating it to the modern React (hooks, context, etc.) as the first milestone.
Later, after "GraphiQL WG" is fully formed and stabilized the group itself can decide to pursue "GraphiQL redesign" as the next big project.

It is just a suggestion, and I don't want to hijack the agenda item. But I want to voice it beforehand to save time during WG and do not deviate discussion from actual "GraphiQL Redesign WG" proposal.

benjie: I'm 💯 with @IvanGoncharov on this; I think we need a dedicated GraphiQL WG, and a redesign can be one of the tasks under that WG.

@AGS- has been performing the GraphiQL merges recently, would be good to get their thoughts and have them on board with this also. @AGS- are you available to attend this WG?

I'm also interested to know what backwards compatibility guarantees we need with GraphiQL: is there a minimum version of Node, GraphQL, React, etc that it must support? Another question for the WG I think :)

orta: Sure, this is all fine with me 👍

leebyron: Excited to have this conversation!

PR #147, opened 2019-04-26
Add Tim, Input Unions to 2019-05-02 agenda

tgriesser:
Added input union RFC to the agenda. Hopefully @binaryseed and @acao will be able to join as they've been dong most of the research work around this topic lately.


acao:
@tgriesser I'll be there!

PR #171, opened 2019-05-09
Add Cory Monty to 2019-06-06 + Dynamic variable declaration

cmonty:
Related to https://github.com/graphql/graphql-spec/issues/377

PR #186, opened 2019-06-05
Remove Cory Monty and agenda item

cmonty:
@leebyron sorry I wasn't able to catch this before merging! We won't have anything to show for our agenda item as we're still working on the suggestions proposed.

Relates to #171


leebyron:
Not a problem! Thanks for the update

PR #192, opened 2019-06-06
Add agenda item to review previous meeting's action items

BrunoScheufler:
The idea of this addition was that it might be a good idea to check on the previous meeting's action items and their respective progress during every meeting so there are no unhandled roadblocks for moving ahead 🚀

### DIFF
--- a/agendas/2019-07-03.md
+++ b/agendas/2019-07-03.md
@@ -39,7 +39,7 @@ Lee Byron            | GraphQL Foundation | Menlo Park, CA
 
 1. Opening of the meeting
 1. Introduction of attendees
-1. Review [previous meeting's action items](../notes/2019-06-06.md) and check progress on those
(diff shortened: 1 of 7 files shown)


agendas/2019-07-03.md
```diff
@@ -39,6 +39,7 @@ Lee Byron            | GraphQL Foundation | Menlo Park, CA
 
 1. Opening of the meeting
 1. Introduction of attendees
+1. Review [previous meeting's action items](../notes/2019-06-06.md) and check progress on those
```
IvanGoncharov: Would be great to link directly to "Action item" section, e.g. https://github.com/graphql/graphql-wg/blob/master/notes/2019-06-06.md#action-items

BrunoScheufler: @IvanGoncharov great idea! I've updated all occurrences to include the anchor 👍

PR #255, opened 2019-10-03
Clarify "Input Union" Agenda item

IvanGoncharov:
I think we now have enough content in the RFC document to review it on WG so we can be sure that everyone on board even if they missed GitHub emails.
So I expanded "Input Union" Agenda item and also added some missing details (length and name of champion). 
@binaryseed What do you think and can you please review proposed changes?


binaryseed:
👍

IvanGoncharov:
@binaryseed Merged 🎉 Thanks for the fast review 👍

PR #258, opened 2019-10-07
Add Craig Smitham to 10/10/19 attendees

craigsmitham:


IvanGoncharov:
@craigsmitham Merged 🎉
BTW, feel free to add items to the agenda (we have only one at the moment).

PR #267, opened 2019-10-09
Add Oscar Funes to the agenda

osukaa:
I want to attend the meeting as a listener. :)


leebyron:
That's fine, I'll add you to the agenda.

However note that we really encourage every attendee to either present an agenda item or be ready to add discussion for an agenda item. Otherwise we publish detailed notes and soon will be publishing recordings of each session

osukaa:
Thank you! I want to start more active participation, but had the curiosity of listening first :)

PR #268, opened 2019-10-09
Add repeatable directives to agenda

spawnia:
The RFC where this change originated from has been around for over a year.

`repeatable` directives have actually been implemented in the reference implementation and are a non-breaking addition that enables use cases that are currently blocked.


IvanGoncharov:
@spawnia Thanks for raising this issue 👍 
It's really sad that we lost Oleg 😢 
Yes, I can definitely give an overview of this proposal especially since I think all reviewed points are already addressed.

PR #281, opened 2019-11-04
Add Benjie to November WG

benjie:
I normally help with the notes; might as well make it official :wink:


IvanGoncharov:
@benjie Thanks for helping with notes 👍
Now it's official 📜

PR #284, opened 2019-11-06
Add Evan and custom scalar RFC

eapache:
Tangent: could we please decide whether we're using two-letter country codes or two-letter state codes? I almost put `Ottawa, CA` until I looked at @leebyron's and realized everyone would think I was in California.


IvanGoncharov:
@eapache Merged 🎉 
An interesting question, I think we can have a quick discussion about it during WG.
My 0.02$ would be to force everyone to add country code at the end `Menlo Park, CA, US`

PR #298, opened 2019-11-07
Swap agenda order

binaryseed:
I'm triple booked at 9am, so I'm shifting the Input Union discussion to later

@IvanGoncharov @leebyron


leebyron:
Not a problem!

PR #306, opened 2019-11-21
Add Rohan as attendee

appwiz:


IvanGoncharov:
@appwiz We had problem with Zoom here is new link: 
https://meet.google.com/gop-aqar-ssv

PR #332, opened 2020-01-07
Add Alan Cha to agenda

Alan-Cha:
Unfortunately, I may miss the first half of the meeting but I will try to participate and take notes when I can.


IvanGoncharov:
@Alan-Cha Thanks for volunteering to take a notes 👍 
Merged 🎉

PR #355, opened 2020-02-05
Add Matt Farmer to 2020-02-06 agenda

m14t:
I'd like to provide a minor update on the progress of the "Custom Scalar Specification URIs" work.

I have one question about the implementation, but it should be fairly quick. If possible to have this early in the meeting to knock it out that would be great.

### DIFF
--- a/agendas/2020-02-06.md
+++ b/agendas/2020-02-06.md
@@ -25,6 +25,7 @@ agenda, edit this file.*
 | Name                     | Organization / Project   | Location
 | ------------------------ | ------------------------ | ------------------------
 | Lee Byron                | GraphQL Foundation       | San Francisco, CA, US
+| Matt Farmer              | Individual Contrib       | Oakland, CA, US
 | *ADD YOUR NAME ABOVE TO ATTEND*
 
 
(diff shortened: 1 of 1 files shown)


eapache:
> Seeking validation on how to handle use-case where a scalar is extended multiple times, which results in the directive being included in the `typeExtensionsMap` multiple times.  Currently taking the last element.

According to https://spec.graphql.org/draft/#sec-Scalar-Extensions we have "Any non‐repeatable directives provided must not already apply to the original Scalar type". I'm not sure if we need to do anything special to define `@specifiedBy` as non-repeatable, but that seems like the right approach?

eapache:
Also, thanks for keeping this going while I was away!

leebyron:
Directives are non-repeatable unless specifically defined to be repeatable. We had to set it up this way to ensure backwards compatibility. All existing directives had to retain the prior behavior of non-repeatability. 

I think the spec text could use an example of a schema definition which breaks this validation rule to help illustrate

m14t:
For history's sake, the comment above was instigated by my PR attempting to handle an invalid use-case.  There was a bug in `graphql-js` which mistakenly allowed repeating non-repeatable directives in extensions.

This was:
- Reproduced with a failing test case in https://github.com/graphql/graphql-js/pull/2440
- Reported in https://github.com/graphql/graphql-js/issues/2442
- Fixed in https://github.com/graphql/graphql-js/pull/2446

PR #379, opened 2020-03-20
Add minutes from Jan, Feb, and Mar 2020 meetings

brianwarner:
I've converted the meeting minutes from Google Docs to markdown.

For what it's worth, I'm using the "Docs to Markdown" Gsuite app, and getting accurate results.

### DIFF
--- a/notes/2020-01-09.md
+++ b/notes/2020-01-09.md
@@ -1,0 +1,195 @@
+# GraphQL Spec WG January 2020
+
+
+### Agenda
+
+
+
+1. Introduction of attendees (5m, Lee)
+2. Determine volunteers for note taking (2m, Lee)
+3. Review agenda (2m, Lee)
+4. Review [previous meeting's action items](https://github.com/graphql/graphql-wg/blob/master/notes/2019-12-05.md#action-items) (5m, Lee)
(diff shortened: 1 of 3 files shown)


benjie:
Nice, good job @brianwarner. Is there any changes we should adopt into note taking to improve the markdown output? I already try and use the header 1/2 presets rather than manual formatting, but I’m wondering if we can do something to make the action highlights come through? Perhaps just bold them in future?

benjie:
Please can someone merge this? @IvanGoncharov  @leebyron

leebyron:
Amazing, thanks again!

brianwarner:
@benjie What you did was great.  Please feel free to keep using the headers, the conversion script picks them up and Does The Right Thing.

As for the action highlights, I might suggest accumulating them in a list at the top of the minutes, or at the end.  To answer the other part of your question, the formatting itself doesn't really matter so long as it's something which translates to markdown (e.g., bold and italic are fine, but text coloring or highlights might not work).

@leebyron You're welcome!  I'll keep an eye on these for you from now on.

benjie:
Thanks @brianwarner :+1:

PR #391, opened 2020-04-15
Add notes from April 2, 2020

brianwarner:
Converted to MarkDown from the live meeting notes.  The original meeting
document is here: https://github.com/graphql/graphql-wg/blob/master/agendas/2020-04-02.md

### DIFF
--- a/notes/2020-04-02.md
+++ b/notes/2020-04-02.md
@@ -1,0 +1,148 @@
+
+# GraphQL WG Notes - April 2020
+
+## Agenda
+
+1. Introduction of attendees (5m, Lee)
+2. Agree to Membership Agreement, Participation Guidelines and Code of Conduct (1m, Lee)
+    *   [Specification Membership Agreement](https://github.com/graphql/foundation)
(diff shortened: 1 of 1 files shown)


notes/2020-04-02.md
```diff
@@ -0,0 +1,148 @@
...
+*   Ivan: I agree with Lee - usual clients don't use introspection in production, so once the app is compiled and released most clients wouldn't notice the difference.
+*   Robert: can we try and document the differences between these different kinds of breaking changes?
+*   Lee: absolutely, that's definitely a gap in the spec.  
+*   Matt M: even some of the "breaking changes" are not necessarily always breaking changes - e.g. changing a type to an interface may not break client queries (but it also might, because __typename may change)
```
benjie, line 59: This broke the highlighting of the raw markdown.

```suggestion
*   Matt M: even some of the "breaking changes" are not necessarily always breaking changes - e.g. changing a type to an interface may not break client queries (but it also might, because `__typename` may change)  
```

brianwarner: Good eyes, thanks @benjie!

leebyron:
Thank you both!

PR #411, opened 2020-05-13
Add Zoom password, update calendar links

brianwarner:
As discussed in the WG meeting last week, the Zoom now has a password.  I've also consolidated the meetings onto one calendar, so that we can more easily manage who is using the Zoom account when.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,7 @@ Pacific US time. Check the [`agendas/`](./agendas) for the exact date and time
 of upcoming meetings.
 
 Keep track of future upcoming meetings by subscribing to the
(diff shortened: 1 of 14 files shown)


benjie:
Might it make sense to remove or strike-through all the "subscribe to the calendar" links for all previous WGs so that people viewing old WG docs don't subscribe to an abandoned calendar?

brianwarner:
That's a good point, @benjie. I've just pushed changes to the other files which referenced the prior calendar or the zoom link.

brianwarner:
I'll go ahead and merge this so that people know how to connect next week.

PR #413, opened 2020-05-14
Add .gitignore

brianwarner:
Ok, this is a trivial one, but it means I won't have to keep unstaging .DS_Store files from my commits :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store


m14t:
FWIW: It's possible to [create a global `.gitignore` file](https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer).

I personally find ignoring these files at the machine level preferable since the `.DS_Store` files are specific to that machine, not this repository. This way I don't need to add this entry to the `.gitignore` file on every repo I work on.

However, I realize that there are a lot of us are using Macs, all of which could run into this issue, and so while it might not be the "most proper" approach, it might be the pragmatic one.

No strong preference from me either way, just offering a potential alternative.

leebyron:
This is the guide I usually point to: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

Agreed that global gitignore is the way to go for OS level noise files, though man does git not make that obvious or easy. I really wish out of the box git installs set up a global ignore for you with your OS's garbage of choice

PR #414, opened 2020-06-01
Add myself and agenda item

eapache:


leebyron:
Before I merge this I want to point out the date change in #417. Please double check your calendar and confirm.

eapache:
👍  the 11th also works for me.

PR #415, opened 2020-06-01
Add Andrew to attendees

asprouse:


leebyron:
Before I merge this I want to point out the date change in #417. Please double check your calendar and confirm.

asprouse:
@leebyron Thanks for the heads up! I will attend on the 11th

PR #416, opened 2020-06-02
Add Benjie to June 2020 WG

benjie:
The "query query query" stuff is progressing, but not enough to add it to the agenda. I'm currently compiling a glossary and trying to determine the best terms to use for things.

I've also started some more spec work in the form of [formatting it with prettier](https://github.com/graphql/graphql-spec/pull/727).


leebyron:
Awesome work Benjie - before I merge this I want to point out the date change in #417. Please double check your calendar and confirm.

benjie:
Oh that’s good actually 🙌 See you on 11th.

PR #472, opened 2020-09-03
Added `dotansimha` to September 2020 WG 

dotansimha:


IvanGoncharov:
@dotansimha Thanks for bringing this up definitely will use help with some stuff.
I will create an issue on the graphql-js repo to discuss time for a call so not only people on this call can participate.
One thing I would suggest is it moves the TS agenda item to be part of the upcoming `graphql-js` call so we don't take the time of people who don't interested in `graphql-js`.

PR #475, opened 2020-09-03
Add Mike Cohen to attendees for 2020-09-03

mcohen75:
A representative from Indeed has signed and executed the Specification Member Agreement.


mcohen75:
Thank you for merging! I've attempted to join the zoom call, but it requires a password. Can you share this with me or point me to where I should get it from?

leebyron:
You'll find it in the agenda - https://github.com/graphql/graphql-wg/blob/master/agendas/2020-09-03.md

Call will start in a couple minutes

PR #482, opened 2020-09-04
Add notes for September 2020 WG

benjie:
Converted via "Docs to Markdown" Gsuite app as suggested by @brianwarner in #379, then piped through `prettier`.

Is it okay to self-merge these in future? It'll make it easier to link them from the action item issues.


leebyron:
Feel free to proactively merge going forward! Thanks for being super timely in getting this up

PR #496, opened 2020-09-14
Add Benjie to October WG

benjie:
Self merging to reduce conflicts.


benjie:
Oops; should have squash merged. Will be more careful next time :+1:

PR #497, opened 2020-09-14
Add query ambiguity topic to October 2020 WG

benjie:
There's a reasonable chance that I won't have stable internet on this date*; if so I'll try and give 24 hours notice and request that we move this topic to November's WG.

\* _moving house soon; and can't go to another location due to COVID-19_.


leebyron:
Good luck with the upcoming move!

PR #553, opened 2020-11-25
Spend some time reviewing previous actions (catch up)

benjie:
@leebyron I propose that we spend a little longer this month going through some of the older action items that I believe can be closed, I think we can close a few of our older ["projects"](https://github.com/graphql/graphql-wg/projects) this way.

I propose that:

- We look at all the items from WGs _before_ July 2020
- We look at newer items that are [marked ready for review](https://github.com/graphql/graphql-wg/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22Ready+for+review+%F0%9F%99%8C%22)

I'm happy to help lead through this.


leebyron:
That sounds great. I may move this to later in the agenda since it's longer this time around

PR #555, opened 2020-12-01
Add Yogesh to Dec meeting

yogesh-desai:


benjie:
Hi @yogesh-desai, is there any project you’re associated with that relates to GraphQL that you could put instead of “Individual”? Doesn’t need to be a company, could be an open source project or a startup or similar. No worries if not.

yogesh-desai:
sure, @benjie added the org name.
Thank you.

benjie (approved):

PR #559, opened 2020-12-03
Add Alan Cha to attendees

Alan-Cha:


Alan-Cha:
I'm so sorry for the late notice. I could have sworn that I made a PR much earlier today. Perhaps I skipped a confirmation step or something.

PR #597, opened 2021-02-02
Add Eloy Durán to Feb 2021 WG

alloy:
I want to start joining regularly, and help out @benjie with note taking, but this first time around I will probably join up to an hour later, if that’s ok?


benjie (approved):

benjie:
Awesome I could definitely do with help on notes! Noted regarding late arrival. So far there's not much on the agenda though.

PR #671, opened 2021-04-30
Move Benjie's May topics to June

benjie:
Personal reasons.

@eapache I've taken the liberty of moving your topic too since it relates to `@oneOf` - as the `@oneOf` champion I'd certainly like to be there (and coherent) for that discussion. I won't merge this without your approval.


eapache (approved):
Of course, hope everything is well!

PR #674, opened 2021-05-04
Add Hugh to the 2020-05-06 meeting

hwillson:
I'm interested in attending this weeks meeting. Thanks very much!

### DIFF
--- a/agendas/2021-05-06.md
+++ b/agendas/2021-05-06.md
@@ -35,6 +35,7 @@ agenda, edit this file.*
 | Benjie Gillam            | Graphile                 | Southampton, UK
 | Rob Van Gennip           | Shopify                  | Toronto, ON, CA
 | Mark Larah               | Yelp                     | San Francisco, CA US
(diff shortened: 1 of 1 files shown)


hwillson:
I'm attending on behalf of Apollo, and I think we're already covered by a corporate CLA. Let me know if not, and I'll get that fixed. Thanks!

brianwarner:
Hi Hugh, yes, Apollo has signed. I just looked, and Apollo's spec membership is configured to cover anyone who is a member of the Apollo GitHub org. I see you're in the group, but you also (probably) would need to click through the above link to verify that this is what you want.

For context, there are two steps when there's a corporate signature. First, the company claims you under their membership, and second, you confirm you want to be claimed. I don't have access to this level of detail in the system, but that would be my first idea. Can you please give that a shot and let me know if it works?

hwillson:
Thanks very much for the context @brianwarner - that worked! 👍

benjie (approved):

PR #676, opened 2021-05-06
POSTPONE MAY WG MEETING

leebyron:


leebyron:
Please take notice @graphql/tsc and current attendees (@lizjakubowski @mjmahone @twof @hwillson @magicmark @ravangen @benjie @eapache)

I'm really sorry for the < 24h notice, but I'm postponing our May WG meeting by one week. I understand if that means some may not be able to join due to schedule conflicts.

PR #772, opened 2021-11-04
Client Controlled Nullability RFC

twof:


benjie (approved):

benjie:
I believe we can just merge this at will since it's an RFC document, but since it's the WG in an hour and a half I'll wait and confirm there :+1:

leebyron:
Thanks for getting this into a document!

twof:
Sure thing!

PR #805, opened 2021-12-02
Add Hugh Willson to the 2021-12-02 agenda and fix meeting date

hwillson:
I would love to attend today!

P.S. > It looks like the date is wrong in the Agenda for today's meeting. I've added a second commit to match it with the Foundation calendar.

Thanks!


benjie:
@hwillson I think the agenda is the "source of truth" and actually the calendar is wrong. The meeting is always 7pm UTC (previously it was always 5pm UTC) but recently (since last meeting) America's clocks changed and the calendar did not adjust. I spotted the same issue and raised it in the chat not even an hour ago: https://discord.com/channels/625400653321076807/862834364718514196/915920541452476417

hwillson:
Thanks @benjie - but the meeting is today right? The agenda doc says it's on Saturday.

benjie:
Ah yes, my apologies - not only is the calendar time wrong but the agenda date is wrong :laughing:

hwillson:
I didn't notice the time being off! 😂 Argh calendaring is hard!

PR #818, opened 2021-12-03
TSC: Add exception to definition of "voting members"

leebyron:
**Charter change:** Adds an exception for elections that all may vote, not just "voting members"

PR #820, opened 2021-12-10
Redefine quorum to allow all to vote

leebyron:
This is an alternative proposal to https://github.com/graphql/graphql-wg/pull/818 which, while more complicated, is a better fit for the goals and principles, namely:

* All TSC members should have the capacity to vote, regardless of what we're voting on.
* TSC votes should remain efficient and resilient to any member not participating.

To do this, I'm proposing changing the way we vote by replacing the concept of a "voting member" with a "quorum":

* A "quorum" is the total number of votes needed for a vote to be considered valid, and this proposes defining that in a similar way we used to define voting members, but *simplifies this to "one of three" instead of "two of five"*.

* Removes the concept of a "voting member" - this means all members are always eligible to vote, even if they do not count towards quorum.

### DIFF
--- a/GraphQL-TSC.md
+++ b/GraphQL-TSC.md
@@ -153,7 +153,13 @@ To provide consistency TSC elections will begin in 2021 for the 2022 term. Prior
 | 2023 - 2024 | Nov 1, 2022      | Dec 1, 2022 | Jan 1, 2023 | Dec 31, 2024 |
 | 2024 - 2025 | Nov 1, 2023      | Dec 1, 2023 | Jan 1, 2024 | Dec 31, 2025 |
 
-### Voting process
+#### Resignation or removal of TSC members
+
+A TSC member may voluntarily resign at any time before the end of their term by notifying the TSC chair. A TSC member may also be involuntarily removed by a supermajority (2/3) vote by TSC members.
+
+After a TSC seat vacancy the remaining TSC will collect self-nominations and vote on a replacement representative to complete the term. The nomination period should be open for at least two weeks, after which a vote shall be held.
+
+## Voting process
 
 When something cannot be decided by consensus or a formal vote is required, the TSC will use a method appropriate to the situation:
 
@@ -162,24 +168,43 @@ When something cannot be decided by consensus or a formal vote is required, the 
 
 The ballots may be public or private depending upon the situation, but the aggregate results should always be published.
 
-#### Minimum threshold for votes
+### Defining a quorum of "attending" members
 
-Because we work in a distributed environment, the voting process must account for a range of time zones and schedules. A quorum of at least half (or supermajority when relevent) of the TSC voting members must cast a ballot in order for the vote to be valid. Once the voting threshold has been met, one of these two critera must be satisfied to conclude the vote:
+To balance preserving the voting ability of all TSC members with the desire for the voting process to remain efficient and avoid deadlocks, we define a valid quorum based on members' recent attendance of GraphQL Working Group meetings.
+
+A quorum is a majority (more than half, or 2/3 for a supermajority vote) of the TSC *attending members*. A quorum must cast a ballot in order for a vote to be valid.
+
+A TSC *attending member* is a member who has attended one of the previous three meetings. Should a TSC member miss three consecutive meetings, they will no longer be counted when determining quorum (but may still vote). A member starts counting towards quorum as of attending a meeting.
+
+### Voting process
+
+Because we work in a distributed environment, the voting process must account for a range of time zones and schedules. Once the threshold of a quorum has been met and a vote is valid, one of these two critera must be satisfied to conclude a vote:
 
 * A notice is sent via email that the vote will conclude in three business days, reminding those who haven't voted that they should do so. The vote will conclude at the end of this time.
-* The election results will not change if additional TSC members vote.
+* The election results would not change with the addition of another vote.
 
-#### Ensuring efficient votes
+Once a valid vote is concluded, the result is determined by the number of votes recieved at that time (as opposed to the total number of TSC members):
 
-A TSC member must attend two of the previous five meetings to be considered a "TSC voting member" and be eligible to vote. Should a TSC member miss three or more of the prior five meetings, they will no longer vote or be counted toward the total number of voting members. This is to ensure broad participation as well as to avoid voting deadlocks.
+* For a simple majority (or supermajority), the votes in favor must exceed half (or 2/3) of the total number of votes.
+* For the Condorcet method, all votes received at the time the vote is concluded are considered.
 
-A member may regain their voting status by attending two meetings and may participate in votes in the second meeting.
+TSC members are not required to vote. An abstention vote is not considered part of the total set of votes, nor towards a voting quorum.
 
-#### Resignation or removal of TSC members
+<details>
+<summary>Example</summary>
 
-A TSC member may voluntarily resign at any time before the end of their term by notifying the TSC chair. A TSC member may also be involuntarily removed by a supermajority (2/3) vote by eligible TSC voting members less any with a conflict of interest.
+Consider this scenario for a simple majority vote:
 
-After a TSC seat vacancy the remaining TSC will collect self-nominations and vote on a replacement representative to complete the term. The nomination period should be open for at least two weeks, after which a vote shall be held.
+* 6 of 11 TSC members are considered *attending* members by having attended one of the last three WG meetings.
+* Being a simple majority, more than half of this 6 must cast votes to consider the vote valid. 4 votes are cast, so the vote is valid.
+* An email is sent out reminding more members to cast a vote within three business days.
+* 1 additional TSC *attending* member casts a vote.
+* 2 members who are not considered an *attending* member also cast a vote.
+* Three days pass, the vote is concluded.
+* A total of 7 votes have been cast.
+* The votes in favor must exceed half of the 7 total votes.
+* At least 4 votes must be in favor for the vote to pass.
+</details>
 
 ## Policies and procedures
 
@@ -195,9 +220,9 @@ Pull requests against this document that do not conflict with the [Technical Cha
 * There are two approvals by TSC members (not including the author)
 * The PR has been open for at least 72 hours
 
-Pull requests that change governance of the TSC (excluding the charter) must be open for at least 14 days, unless consensus is reached in a meeting with quorum of TSC voting members.
+Pull requests that change governance of the TSC (excluding the charter) must be open for at least 14 days, unless consensus is reached in a meeting with quorum of TSC *attending* members.
 
-If consensus cannot be reached, a pull request may still be landed after a vote by the TSC voting members to override outstanding objections.
+If consensus cannot be reached, a pull request may still be landed after a vote by TSC members to override outstanding objections.
 
 An exception is made for errata or to update meeting logistics. These may be landed immediately, provided all EasyCLA checks have passed.
 
--- a/GraphQL-TSC.md
+++ b/GraphQL-TSC.md
@@ -193,10 +193,29 @@ Once a valid vote is concluded, the result is determined by the number of votes 
 TSC members are not required to vote. There are three ways an *attending* member may reply to choose not to vote, each with a different intent and impact on the voting process:
 
 * **Present:** A reply of "present" causes an *attending* member to count towards quorum, but does not count towards the number of votes. This has the practical impact of allowing progress towards reaching quorum while lowering the number of "yes" votes required for a motion to pass. A member may reply "present" if they support a motion reaching a result, but do not have a preference if it passes or fails.
-* **Absent/Abstain:** A reply of "abstention" causes an *attending* member to neither count towards quorum, nor towards the number of votes. This is implied for any absent *attending* member which does not reply to a vote. The has the practical impact of hindering progress towards reaching quorum. A member may reply "abstain" if they do not support a motion reaching either result.
+* **Absent/Abstain:** A reply of "abstention" causes an *attending* member to neither count towards quorum, nor towards the number of votes. This is implied for any absent *attending* member which does not reply to a vote. This has the practical impact of hindering progress towards reaching quorum. A member may reply "abstain" if they do not support a motion reaching either result.
 * **Recused:** A reply of "recusal" causes a member to not count as an *attending* member for the purposes of this vote. This has the practical impact of both lowering the number of *attending* members required to reach quorum, and the number of "yes" votes required for a motion to pass. A member may reply "recused" if they want to remove themselves to minimize their impact on a vote, often due to a conflict of interest.
 
 Note: Non-attending members do not count towards quorum, and should they not vote are effectively recused.
+
+### Voting procedure
+
+Bringing this all together, here's a step-by-step procedure for administering a TSC vote:
+
+1. Formally open a vote for a motion by opening a GitHub issue.
+   * Most votes are held publicly, for those held privately, open an issue in a private forum.
+1. Take attendence of all TSC members over the prior three meetings to determine *attending members*.
+   * If a vote is raised during a meeting, count the live meeting as the most recent of the three.
+1. Determine the *quorum threshold* as the next number larger than half of *attending members*, or 2/3 for a supermajority.
+   * Examples: For 8 attending, a quorum threshold is 5. For 11 attending, quorum threshold is 6. For 9 attending, a supermajority quorum threshold is 7.
+1. Gather votes from *all members* until the number of votes from *attending members* reaches *quorum threshold*.
+   * A reply of "present" or "recused" may require recalculating *attending members* and/or *quorum threshold*.
+   * Send recurring reminders to remaining *attending members*.
+1. Message that the vote will be closed in 3 business days, make good faith attempt at reaching out to all remaining members.
+   * Conclude the vote early if, for single-winner votes, 6 *yes votes* or 5 *no votes* are received for a majority vote, or 8 *yes votes* or 3 *no votes* for a supermajority, as no additional set of votes could change the outcome.
+   * Any "present" or "recused" replies lower the thresholds for early conclusion.
+1. The resulting motion passes if the number of *yes votes* is a majority (or 2/3 supermajority) of *vote count*.
+1. Should a vote take place in a private forum, post the result in a public forum.
 
 <details>
 <summary>Example</summary>
--- a/GraphQL-TSC.md
+++ b/GraphQL-TSC.md
@@ -139,7 +139,7 @@ The [Technical Charter](https://github.com/graphql/foundation/blob/master/GraphQ
 
 ### Election process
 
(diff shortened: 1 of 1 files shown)


GraphQL-TSC.md
```diff
@@ -153,7 +153,13 @@ To provide consistency TSC elections will begin in 2021 for the 2022 term. Prior
 | 2023 - 2024 | Nov 1, 2022      | Dec 1, 2022 | Jan 1, 2023 | Dec 31, 2024 |
 | 2024 - 2025 | Nov 1, 2023      | Dec 1, 2023 | Jan 1, 2024 | Dec 31, 2025 |
 
-### Voting process
+#### Resignation or removal of TSC members
```
leebyron, line 156: This is just a move to keep this section together and focused

nyteshade (approved):
I don’t have any issues with these changes. Looks good to me.

benjie (approved):

GraphQL-TSC.md
```diff
@@ -162,24 +168,44 @@ When something cannot be decided by consensus or a formal vote is required, the
...
+Consider this scenario for a simple majority vote:
+
+* 6 of 11 TSC members are considered *attending* members by having attended at least two of the last five meetings.
+* Being a simple majority, more than half of this 6 (so 4 or more) must cast ballots to consider the vote valid. 4 ballots are cast.
+* An email is sent out reminding more members to cast a vote within three business days.
+* 1 additional TSC *attending* member casts a vote.
+* 2 members who are not considered an *attending* member also cast a vote.
+* The vote is concluded.
+* A total of 7 ballots have been cast
+* A majority is determined relative to the number of *attending* members (6) plus the number of non-attending but voting members (2), totaling 8.
```
mjmahone, line 205: I don't know why this was confusing to me on first read, but given the numbers provided I was able to reverse engineer the algorithm to mean what it says (the base 100% number used to evaluate whether majority is reached is all attending members, voting or not, plus any voting but not attending member).

It's in the example so it's not required to be editorially perfect, and also after reading 2-3 times it's very clear and I can't un-realize again to figure out what wording would be more clear without muddying the example.

mjmahone (approved):

GraphQL-TSC.md
```diff
@@ -162,24 +168,44 @@ When something cannot be decided by consensus or a formal vote is required, the
...
+* The election results would not change with the addition of another TSC member's vote.
 
-A TSC member may voluntarily resign at any time before the end of their term by notifying the TSC chair. A TSC member may also be involuntarily removed by a supermajority (2/3) vote by eligible TSC voting members less any with a conflict of interest.
+Once a valid vote is concluded, the vote outcome is determined as follows:
 
-After a TSC seat vacancy the remaining TSC will collect self-nominations and vote on a replacement representative to complete the term. The nomination period should be open for at least two weeks, after which a vote shall be held.
+* For a simple majority (or supermajority), the votes in favor must exceed 50% (or 66%) of the total of, as of the time the vote is concluded:
+  * the number of TSC *attending* members.
+  * the number of members who are not considered a TSC *attending* member but did cast a vote.
+* For the Condorcet method, all votes received at the time the vote is concluded are considered.
```
robzhu, line 191: I had trouble understand this paragraph, would something like this rephrasing work:

> A simple majority requires more than 50% of the TSC attending members to vote in favor. 

I don't recall us referencing the Cordorcet method elsewhere. If this is the first/only time referencing the Condorcet method, considering in-lining the relevant parts and eliding the term altogether.

robzhu (approved):

GraphQL-TSC.md
mjmahone: But more than 50% of the attending members isn’t quite right. 

It’s:
- take the sum of attending members and non-attending members who voted
- Majority is based on that sum, so >50% of whatever the number in the first step.

GraphQL-TSC.md
benjie: I had the same process.

GraphQL-TSC.md
leebyron: Cordorcet has been in here since the beginning, it's referenced two headings up - we use it for multiple-winner votes (eg elections)

Let me see if I can refine the explanation of this - I agree that it's hard to reason about

GraphQL-TSC.md
leebyron: Yeah :/ I think I'm missing a defining term for this

GraphQL-TSC.md
robzhu: Looks like I need more coffee, thanks guys, sorry for the noise, no other comments from me.

leebyron:
Ok @graphql/tsc - good feedback, I've simplified this by counting a majority relative to *votes* rather than the harder to explain union of attending/voting. This has a minor downside of potentially allowing a vote by a non-representative majority, but has the benefits of being way easier to explain and aligns well with the goal of keeping elections efficient and avoiding deadlocks.

leebyron:
@graphql/tsc please take another look since this is materially different (and hopefully simpler/better)

GraphQL-TSC.md
```diff
@@ -162,24 +168,43 @@ When something cannot be decided by consensus or a formal vote is required, the
 
 The ballots may be public or private depending upon the situation, but the aggregate results should always be published.
 
-#### Minimum threshold for votes
+### Defining a quorum of "attending" members
+
+To balance preserving the voting ability of all TSC members with the desire for the voting process to remain efficient and avoid deadlocks, we define a valid quorum based on members' recent attendance of GraphQL Working Group meetings.
 
-Because we work in a distributed environment, the voting process must account for a range of time zones and schedules. A quorum of at least half (or supermajority when relevent) of the TSC voting members must cast a ballot in order for the vote to be valid. Once the voting threshold has been met, one of these two critera must be satisfied to conclude the vote:
+A quorum is a majority (more than half, or 2/3 for a supermajority vote) of the TSC *attending members*. A quorum must cast a ballot in order for a vote to be valid.
```
mjmahone, line 175: This seems to explicitly exclude non-attending members from being able to count towards quorum. So if you have 6 attending members even if you can get 5 non-attending members and 3 attending to vote, you still haven’t achieved quorum. 

I think that probably matches the spirit of what we’re going for, so seems good to me.

GraphQL-TSC.md
```diff
@@ -162,24 +168,43 @@ When something cannot be decided by consensus or a formal vote is required, the
...
+
+A TSC *attending member* is a member who has attended one of the previous three meetings. Should a TSC member miss three consecutive meetings, they will no longer be counted when determining quorum (but may still vote). A member starts counting towards quorum as of attending a meeting.
+
+### Voting process
+
+Because we work in a distributed environment, the voting process must account for a range of time zones and schedules. Once the threshold of a quorum has been met and a vote is valid, one of these two critera must be satisfied to conclude a vote:
 
 * A notice is sent via email that the vote will conclude in three business days, reminding those who haven't voted that they should do so. The vote will conclude at the end of this time.
-* The election results will not change if additional TSC members vote.
+* The election results would not change with the addition of another vote.
```
mjmahone, line 184: I’m not sure this is what we want: this says if the results are 5-3 for a majority vote, then the vote is concluded. I think instead of “another” we need “the election results would be the same even if every not-yet-voting member cast the opposite vote”. Or “results could not change with the addition of other votes”

benjie, line 184: I’m not comfortable with this, why is there a one vote threshold? Maybe 
```suggestion
* The election results would not change if all remaining TSC *attending members* (if any) were to vote.
```

GraphQL-TSC.md
```diff
@@ -162,24 +168,43 @@ When something cannot be decided by consensus or a formal vote is required, the
...
 
-#### Ensuring efficient votes
+Once a valid vote is concluded, the result is determined by the number of votes recieved at that time (as opposed to the total number of TSC members):
 
-A TSC member must attend two of the previous five meetings to be considered a "TSC voting member" and be eligible to vote. Should a TSC member miss three or more of the prior five meetings, they will no longer vote or be counted toward the total number of voting members. This is to ensure broad participation as well as to avoid voting deadlocks.
+* For a simple majority (or supermajority), the votes in favor must exceed half (or 2/3) of the total number of votes.
+* For the Condorcet method, all votes received at the time the vote is concluded are considered.
 
-A member may regain their voting status by attending two meetings and may participate in votes in the second meeting.
+TSC members are not required to vote. An abstention vote is not considered part of the total set of votes, nor towards a voting quorum.
```
mjmahone, line 191: Hmmm not sure this makes sense. Usually an explicit vote of abstention counts towards quorum (you’re saying “I was present, understood the argument, and do not care which way the result goes”) whereas an abstention via not voting doesn’t count towards quorum (as that’s saying you may not have had time to figure out what makes sense). We may want to have both a “present” and “abstain-do-not-understand-yet” vote.

mjmahone (commented):
As an example of my comment on abstentions, as I am not yet approving this version, but also not explicitly saying “present” or “either works”, my comment here should fall under the “not for quorum” case, whereas if I said “I read through and am fine either way” that should count towards quorum, IMO.

GraphQL-TSC.md
```diff
@@ -162,24 +168,43 @@ When something cannot be decided by consensus or a formal vote is required, the
...
+### Voting process
+
+Because we work in a distributed environment, the voting process must account for a range of time zones and schedules. Once the threshold of a quorum has been met and a vote is valid, one of these two critera must be satisfied to conclude a vote:
 
 * A notice is sent via email that the vote will conclude in three business days, reminding those who haven't voted that they should do so. The vote will conclude at the end of this time.
-* The election results will not change if additional TSC members vote.
+* The election results would not change with the addition of another vote.
 
-#### Ensuring efficient votes
+Once a valid vote is concluded, the result is determined by the number of votes recieved at that time (as opposed to the total number of TSC members):
```
benjie, line 186:
```suggestion
Once a valid vote is concluded, the result is determined by the number of votes received at that time (as opposed to the total number of TSC members):
```

GraphQL-TSC.md
benjie, line 191: I don’t feel this is quite right; this should be a little analogous to a “conflict of interest” situation - i.e. the quorum should be recalculated as if the abstaining member were not an attending member. This reads as if they don’t count towards reaching the quorum but without also (potentially) reducing the quorum.

We should also have a rule for the minimum size of a quorum; e.g. 2 people.

benjie (changes requested):

GraphQL-TSC.md
leebyron: I think your intent is the same as what's here. We could make it more clear.

Eg for a majority vote, as soon as you get 6 yes votes (of 11 TSC) then there's no need to wait any longer for more people to vote. 

However you wouldn't want this to be based on attending membership since you want to allow room for others to vote. 

In other words, immediately cease collecting votes as soon as the next vote could not possibly affect the result

GraphQL-TSC.md
leebyron: I don't know that I follow your concern. My thought was that an abstain can't simply count as a vote that's not yes, since that would make it equivalent to a no. So it needs to not count as a vote at all such that "a majority of votes" still holds. Also, if someone abstains, that shouldn't trigger a vote quorum either. You need a quorum of yes/no votes to make a vote valid. 

Is there more to this that I've missed?


I think I like the idea of a minimum quorum, but wouldn't that only happen in the case of extreme absence, in which case it might be possible minimum quorum can't be reached and things deadlock?

leebyron: Since quorum is "more than half of attending" then for cases of very low attendance: 4 attending means needing 3 for quorum. 3 attending means 2 for quorum. 2 attending also means 2 for quorum. 1 attending means 1 for quorum.

GraphQL-TSC.md
leebyron: Yeah, you guys are expressing similar concern - but both your sentiments is the intended rule. I'll clarify

GraphQL-TSC.md
leebyron: I think my worry is that if an abstain counts as a vote but we consider a passing result for majority based on yesVotes/allVotes then an abstain is equivalent to a no vote.

I think the only way you can make an abstain not affect the outcome is by also removing it from the denominator.

But if you had a bunch of people abstain from voting, would you still want to consider those towards meeting quorum for considering a vote valid? I suppose that's what a "present" vote would imply? Do we want to support that or should we require yes/no votes for measures to pass?

mjmahone: I would be OK with any of:
- an explicit abstention being impossible (i.e. the vote continues until quorum is reached and the deadline to keep voting passes)
- abstentions counting towards quorum but reducing the vote’s denominator for majority required
- abstention reducing the denominator without adding to the numerator both for quorum and also majority

An abstention not counting towards quorum is equivalent to the first, except it makes it more explicit that if you say you’re abstaining, your vote might still actually be required to close out the issue. Especially if there were 2 attending TSC members, you don’t want one of them to feel like it’s OK to abstain if their vote is in fact required to come to a decision: both of them really do need to keep digging into the issue until they have an explicit opinion, or we need to let one of them completely defer the option to the other to enable a quorum from both with a 1-vote majority.

mjmahone: Hmm taking those three examples out, if we had 3 attending members, then:
- in the first, you always need 2 to vote up or down for there to be quorum. If one votes yes and one no, it fails.
- In the second, you can have one person vote yes and one abstain. Quorum is achieved (2/3), and it will fail (1/2). Whereas if you have two people abstain and one vote yes, it will have quorum (3/3) and pass (1/1). 
- in the third, one person abstaining with one yes vote is not enough to reach quorum (1/2), while two abstaining and one voting achieves quorum (1/1) and still lets the one vote pass it (1/1).

GraphQL-TSC.md
benjie: Right; but imagine you have:

```
Yes: 4
No: 2
```

The addition of another vote (`no`) would not change the result (4 yes/3 no); however the addition of _three_ more votes (three "no"s) could change the result: (4 yes/5 no).

My issue was with "another" taken to mean a singular vote.

I'm totally happy with `if all remaining members were to vote` rather than just the attending members, I was just guessing at your intent.

GraphQL-TSC.md
benjie: TL;DR: "nor towards a voting quorum" is ambiguous - does it mean that they would not count when calculating the quorum, or when calculating if the quorum was fulfilled?

Laying out a concrete example to explain my concern:

7 attending members.

Quorum is therefore 4.

4 attending members abstain.

Three remaining attending members vote (two for, one against).

No non-attending members vote.

You state abstentions are not considered "towards a voting quorum" so it's clear there's only 3 votes that count towards the quorum. However the quorum requirement is 4, therefore this vote is invalid due to too many abstentions? Effectively the abstention is a "no".

Therefore, I think that abstentions should be subtracted from the size of the pool used to calculate the quorum in the first place - 7 attending members less 4 abstentions is 3 remaining attending members. Quorum is thus 2, so the vote is valid, and the result (2 for, 1 against) is that the vote passes. This is similar to how someone would be removed from the quorum-determining pool in the case of conflict-of-interest.

benjie: > wouldn't that only happen in the case of extreme absence, in which case it might be possible minimum quorum can't be reached and things deadlock?

Yes, I'd prefer a deadlock over a single person (out of 11!) getting the one and only vote on the subject. I think that at least two people should have to vote for a change for the change to occur - "favour no change," right? :wink:

The only exception to this would be voting on new TSC members - should we all attend a conference and get hit by a meteor (and thus unable to vote) then the sole remaining TSC member should be able to appoint a new committee such that votes on other subjects may continue.

leebyron:
Thanks @mjmahone @benjie - I just added a commit addressing both of your feedback.

* I used @benjie's suggestion of "if all remaining members were to vote" to clarify intent on when a vote concludes.
* I expanded on the section of non-votes to capture all of the variants described above. No need to restrict ourselves to only one definition - I think it's best to just have a shared set of definitions for these so they're not surprising in the case we have this

GraphQL-TSC.md
leebyron: Hopefully my last edit navigates these concerns.

I added definitions for present/abstain/recused and their impact on quorum that way people can have an informed impact on the process. I think this avoids the concern with the size of quorum dropping to a very low number, even 1. "Abstain" (or no reply) does not allow a vote to occur since the threshold remains the same but hinders progress towards it, where present/recused would either make progress towards that threshold or lower the threshold accordingly - but those are intentionally provided replies, not implied.

The only scenario in which the threshold for quorum can reach 1 in an unintentional way is if all but one member have not attended 3 meetings. That effectively gives the remaining person full power, and I think this is a good thing. This is the hit-by-meteor scenario.

mjmahone (approved):
I think this is really well worded now, and is good documentation for new TSC members on what exactly is available for them: what their obligations are and what actions have which consequences. Thank you!

brianwarner (approved):
This is a novel way to approach this, but that's not a bad thing. I can appreciate that it allows groups to get decisions made quickly at small scale, but ensures TSC members can weigh in if needed.

Thinking about this from the shoes of someone who would be administering a vote, here's how I understand the process:

1. How many TSC members count as "attending" members?
2. What is the threshold of "attending" member ballots required to render this vote valid? (e.g., majority or supermajority)
3. Have enough "attending" members voted to reach that threshold?
    1. If not, which "attending" members do we need to remind?
4. Once the threshold of "attending" members has been met, could any additional votes change the outcome?
    1. If no, conclude the election.
    2. If yes, follow the procedures and deadlines outlined here.

LGTM.

leebyron:
Exactly. In fact, I'll add these admin steps!

brianwarner (approved):
Adding an explicit followup to the recent changes, these LGTM as well.

mjmahone (approved):

nyteshade (approved):

schrockn:
👍🏻

dschafer:
Looks good to me.

leebyron:
Thanks for the careful read and great feedback everyone. If anyone spots any additional grammar/spelling/clarity improvements I'm happy to address those in a follow up

IvanGoncharov (commented):
LGMT 👍

schrockn (commented):
Approved!

benjie (commented):
This looks solid in general, with the one caveat that it's not explicit what would happen if quorum were not reached during a member election vote. I'm assuming the seats would remain empty until a "valid" vote (a vote reaching quorum) takes place. This doesn't seem ideal, but I also don't think it's particularly likely, so I'm okay with it.

Approved.

leebyron:
Yeah the two scenarios where a vote is not valid are:

Either: not enough people respond, in which case what you described is right, we pester people to vote to try not to create a off schedule problem, but not the end of the world if that happens. 

or, you get enough intentional abstains that you cant actually reach quorum. In this case the vote dies on the vine. But there's a distinction between "we voted on this in the past and decided not to do this" vs "we tried to vote on this in the past, but didn't reach a conclusion at that time"

PR #822, opened 2021-12-13
Add Deprecation of input values to January 2022 agenda

fotoetienne:
Also add Stephen to January attendees


benjie:
I’d like to say 30 minutes is probably too long, but… 😬

PR #826, opened 2021-12-17
Add Benjie and security process (TSC Discussion) to January WG

benjie:
@leebyron How would you like the TSC agenda added? I've put it in its own section which I believe makes sense, but if you want it mixed into the normal agenda let me know at which point to add it (I'd guess just before or just after the introduction of attendees?).

I may no longer be on the TSC by the time of this meeting (depending on the election results), but am happy to present this to the TSC or to let a TSC member do so nonetheless :+1:


benjie:
I'm going to go ahead and merge this to avoid (more) conflicts.

leebyron:
Sorry for missing this! Glad to see this on the agenda.

In the spirit of having a linear agenda, I think what we should do for something like this is keep TSC specific agenda items inline with the rest of the agenda, but add a `**TSC:**` prefix so that it's easy to scan the agenda and see what is specifically meant for TSC engagement

PR #839, opened 2022-01-05
New RFC for Fragment Modularity problem + solution space exploration

mjmahone:
There's been a lot of discussion on https://github.com/graphql/graphql-spec/pull/895 around how we're lacking a good solution to error boundaries, which seems to stem primarily from a lack of fragment modularity (in a response, you can't know which fragments are actually a part of it).

After some iterations with @captbaritone and @josephsavona , I've laid out as much of the problem and solution space in this area that I'm aware of. I think we need to act with haste to resolve this key language inadequacy, and I want to do my part to drive some solution in this space forwards.

### DIFF
--- a/rfcs/FragmentModularity.md
+++ b/rfcs/FragmentModularity.md
@@ -1,0 +1,259 @@
+# RFC: Fragment Modularity
+
+**Proposed by:**
+- [Matt Mahoney](https://github.com/mjmahone) - Facebook
+
+**Contributors:**
+- [Joe Savona](https://github.com/josephsavona) - Facebook
+- [Jordan Eldredge](https://github.com/captbaritone) - Facebook
+- You?
+
+GraphQL was originally designed to support building mobile applications, where for performance reasons applications need to fetch the minimum amount of data needed for a given UI surface and aggregate requests into fewer network requests. In parallel with GraphQL’s evolution, component-oriented UI frameworks (ie React) gained traction, supporting a style in which applications could be broken down into modular, reusable components. GraphQL fragments were a natural way for components to describe their data dependencies in a *mostly* modular and reusable way, and Relay and other GraphQL clients have taken this approach to its logical extreme and use fragments as a fundamental unit in their APIs and architectures.
+
+## Problem Statement
+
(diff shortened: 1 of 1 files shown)


josephsavona:
[Rendered version](https://github.com/graphql/graphql-wg/blob/2e5b1798b1c34365f5afe50580b3cb106595b64f/rfcs/FragmentModularity.md)

aprilrd:
Thank you for the write up. I was skeptical of the need for the fragment modularity before reading this, but it changed my mind. 

I prefer options that allow opt-ins and using new syntax as it allows the community to adopt gradually, and doesn't change the current behavior too radically. Of those, "Syntax #1: aliased spreads" is my favorite, followed by #2 and #3.

benjie:
RFC documents get merged quickly; feel free to continue the discussion here or to raise follow-up PRs with suggested changes 👍

eneko:
Hi folks, this is a very interesting proposal. One observation, it seems the JSON example uses mixed Pascal and camel cases for property names:
```json
 {
   "Bar": {
     "account_name": "mjmahone"
   },
   "HasName": {
     "name": "Matt"
   },
   "id": 0
 }
 ```

Ideally, these property names would match the rest of the document. From the suggested syntax options, it seems options 1 & 2 would allow for this, as long as the following was possible.

#### Syntax 1 with `camelCase`:
```graphql
 fragment Foo on Actor {
   bar: ...Bar
   hasName: ... on HasName {
     name
   }
   id
 }
 fragment Bar on HasAccount {
   account_name
 }
 ```

#### Syntax 2 with `camelCase`:
```graphql
 fragment Foo on Actor {
   bar: {
     ...Bar
   }
   hasName: {
     ... on HasName {
       name
     }
   }
   name
 }
 fragment Bar on HasAccount {
  account_name
 }
 ```

Both of the above examples would generate the following JSON:
```json
 {
   "bar": {
     "account_name": "mjmahone"
   },
   "hasName": {
     "name": "Matt"
   },
   "id": 0
 }
 ```

It does not seem Syntax 3 would support camelCase, unless it automatically formatted the output, always enforcing it.

PR #845, opened 2022-01-06
Adjust time budgets for Jan 2022 WG

benjie:
We're massively over-subscribed for our 2 hours window; I've had to free up 40 minutes! To do so, I have made the following changes

1. Assumed that reviewing previous meetings action items should be limited to the single "ready to review" item
2. I've reduced the time allocated to @yaacovCR, @fotoetienne, @ilslv, @robrichard and myself
3. I've moved the shorter items to the front of the agenda

We're going to need to watch the clock and try and avoid going over the time budgets, we can always add more discussion to next months agenda or via PRs/issues/Discord.

@yaacovCR Since your change is two sentences I hope 10 minutes should cover it, and any ongoing discussion can take place on the PR.

@fotoetienne I'm sorry but I had to cut a full 10 minutes from your allocation, make sure you know what outcomes you want from your time and focus on reaching those :+1: 

@ilslv We don't have long enough for the kind of discussion we would normally have, I think you should quickly surface the inconsistency, and if a consensus is not reached rapidly you should collect interested parties (GitHub handles are in the agenda) and then open a discussion via a GitHub issues with those parties to discuss the direction.

@robrichard I had to cut 10 minutes from your allocation too, but since you're at the end if interested parties want to stay longer I guess they can :)

You are all welcome to reach out to me publicly or privately (e.g. via the GraphQL Discord) if you need any help/advice on what to do w.r.t taking discussions offline.

cc @leebyron

PR #848, opened 2022-01-06
Add Ivan to Jan2022 WG agenda

IvanGoncharov:


benjie:
@IvanGoncharov I've moved this item, reduced the time for it, and taken time from other topics to make space. See #845 for more details.

Issue #765, opened 2021-10-28
Technical Steering Committee 2022 Elections


leebyron:
The GraphQL technical project is governed by our [technical steering committee (TSC)](https://github.com/graphql/graphql-wg/blob/main/GraphQL-TSC.md).

The TSC was initially seeded in Nov 2020 with ten individuals plus the exec director (myself) with terms randomly assigned to be 1 or 2 years. Starting this year and going forward, we will hold elections for five seats per year with each elected member to serve a 2 year term.

### TSC election process overview:

Please read the complete [rules for this elections process](https://github.com/graphql/graphql-wg/blob/main/GraphQL-TSC.md#becoming-a-tsc-voting-member). However here is a brief overview:

1. Five TSC members end their terms on Dec 31st.
2. Between now and the end of the year we will nominate and elect five members for these seats.
3. Existing TSC members may (but are not required to) nominate themselves for an additional term.
4. The TSC members not up for election (remaining 5 plus myself) vote on nominees using the [Condorcet method](https://en.wikipedia.org/wiki/Condorcet_method).

### Upcoming dates:

|                  |       |
|  --------- | --- |
| **Nov 1** | Nominations open. If interested, please submit your name with [this form](https://tsc-nomination.graphql.org/).
| **Nov 29** | Nominations close.
| **Dec 1** | Voting by remaining TSC to begin.
| **Dec 27** | Voting ends and new TSC members announced (or earlier).
| **Dec 31** | Prior TSC members' term ends.
| **Jan 1, 2022** | New TSC members' two-year term begins. (Note: The next TSC/WG meeting to be held Jan 6th, 2022)
| **Dec 31, 2023** | New TSC members' two-year term ends.

### Expiring TSC member terms:

| Name                                               | Affiliation           | Term begins | Term ends    |
| -------------------------------------------------- | --------------------- | ----------- | ------------ |
| [Benjie Gillam](https://github.com/benjie)         | Graphile              | Nov 1, 2020 | Dec 31, 2021 |
| [Brielle Harrison](https://github.com/nyteshade)   | PayPal                | Nov 1, 2020 | Dec 31, 2021 |
| [Ivan Goncharov](https://github.com/IvanGoncharov) | Apollo                | Nov 1, 2020 | Dec 31, 2021 |
| [Matt Mahoney](https://github.com/mjmahone)        | Facebook              | Nov 1, 2020 | Dec 31, 2021 |
[...]

PR #856, opened 2022-01-06
Update TSC member list, add emeriti

leebyron:
Based on the results shared in #765


leebyron:
Just a note that I left the Election dates table as is. I find it useful to look back at the dates of the last election, and the next two rows show election dates relevant to all current seats (those ending in Dec 2022 and 2023).

After the next election, we should keep a habit of removing the oldest row and adding a newer row in that table.

benjie (approved):

mjmahone (approved):

michaelstaib (approved):

PR #889, opened 2022-02-14
Adding myself for attendance of 2022 meetings.

Adron:
Did a bulk add for 2022 meetings, hope that's ok. Can submit individually if that is preferred.


Adron:
Can somebody just kick the CLA check for me? I've signed it already and @robzhu has already ok'd it. The EasyCLA site also shows that it has no requests/pending verification.  🤷🏼‍♂️

benjie:
@Adron it looks like the author of that commit is not compatible with what EasyCLA is expecting; did you do something atypical?

Adron:
Ah, just noticed what I did... will resubmit a PR w/o the tertiary author. I just forked, then made the changes in a branch, the merged and submitted *that* PR which altered the author so it doesn't look like I edited the files. At least, it appears. 🤷‍♂️

PR #1006, opened 2022-06-03
Add OptInFeatures.md

martinbonnin:
**[Rendered proposal](https://github.com/martinbonnin/graphql-wg/blob/opt-in-features/rfcs/OptInFeatures.md)**

This is a work in progress centralising the latest state of https://github.com/graphql/graphql-spec/issues/943 in a more reader-friendly way than the Github thread. 

Comments are welcome here or at https://github.com/graphql/graphql-spec/issues/943

### DIFF
--- a/rfcs/OptInFeatures.md
+++ b/rfcs/OptInFeatures.md
@@ -6,8 +6,6 @@
 
 ## 📜 Problem Statement
 
-  
-
 GraphQL has a [built-in mechanism for deprecation](https://spec.graphql.org/draft/#sec--deprecated) allowing to gracefully remove features from the schema. The lifecycle of a feature can typically be represented as `stable` -> `deprecated` -> `removed`.  
 
 In a lot of cases though, a feature lifecycle includes an experimental phase where it has just been added and can be changed without warning. In this state, the feature is usable and feedback is encouraged but isn't considered stable enough to be put in production. The feature lifecycle becomes `experimental` -> `stable` -> `deprecated` -> `removed` .
@@ -31,19 +29,20 @@ This proposal is not about security and/or hiding parts of a schema. Its goal is
 
 * GitHub uses [schema previews](https://docs.github.com/en/graphql/overview/schema-previews) to opt-in new features.
 * Kotlin has [OptIn requirements](https://kotlinlang.org/docs/opt-in-requirements.html) that started out as `@Experimental` before [being changed to `@OptIn`](https://youtrack.jetbrains.com/issue/KT-26216/Generalize-Experimental-API)
+* Atlassian has [a `@beta` directive](https://developer.atlassian.com/platform/atlassian-graphql-api/graphql/#schema-changes) that is enforced during execution. A client must provide a `X-ExperimentalApi: $Feature` HTTP header or the request will fail.
 
 ## 🧑‍💻 Proposed solution
 
-### The `@optIn` directive
+### The `@requiresOptIn` directive
 
-It is proposed to add an `@optIn` directive to the specification:
+It is proposed to add an `@requiresOptIn` directive to the specification:
 
 ```graphql
 """
 Indicates that the given field, argument, input field or enum value requires
 giving explicit consent before being used.
 """
-directive @optIn(feature: String!) repeatable
+directive @requiresOptIn(feature: String!) repeatable
     on FIELD_DEFINITION 
     | ARGUMENT_DEFINITION 
     | INPUT_FIELD_DEFINITION 
@@ -57,8 +56,8 @@ type Session {
   id: ID!
   title: String!
   # [...]
-  startInstant: Instant @optIn(feature: "experimentalInstantApi")
-  endInstant: Instant @optIn(feature: "experimentalInstantApi")
+  startInstant: Instant @requiresOptIn(feature: "experimentalInstantApi")
+  endInstant: Instant @requiresOptIn(feature: "experimentalInstantApi")
 }
 ```
 
@@ -66,7 +65,7 @@ type Session {
 
 > This section is a proposal based on the current introspection mechanism. A more global mechanism (see [#300](https://github.com/graphql/graphql-spec/issues/300)) would make it obsolete
 
-`@optIn` features should be hidden from introspection by default and include if `includeOptIn` contains the given feature:
+`@requiresOptIn` features should be hidden from introspection by default and include if `includeOptIn` contains the given feature:
 
 ```graphql
 type __Type {
@@ -80,7 +79,7 @@ type __Type {
 }
 ```
 
-Tools can get a list of `@optIn` features required to use a field (or input field, argument, enum value) using `requiresOptIn`:
+Tools can get a list of `@requiresOptIn` features required to use a field (or input field, argument, enum value) using `requiresOptIn`:
 ```graphql
 type __Field {
   name: String!
@@ -88,7 +87,7 @@ type __Field {
 
   # [...] other fields omitted for clarity
 
-  # list of @optIn features required to use this field
+  # list of @requiresOptIn features required to use this field
   requiresOptIn: [String!]
   args(includeDeprecated: Boolean = false, includeOptIn: [String!]): [__InputValue!]!
 }
@@ -102,9 +101,9 @@ includeOptIn.containsAll(field.requiresOptIn) && (includeDeprecated || !field.is
 
 ### Validation
 
-Similarly to [deprecation](https://spec.graphql.org/draft/#sel-FAHnBZNCAACCwDqvK), the `@optIn` directive must not appear on required (non-null without a default) arguments or input object field definitions.
+Similarly to [deprecation](https://spec.graphql.org/draft/#sel-FAHnBZNCAACCwDqvK), the `@requiresOptIn` directive must not appear on required (non-null without a default) arguments or input object field definitions.
 
-In other words, `@optIn`  arguments or input fields, must be either nullable or have a default value.
+In other words, `@requiresOptIn`  arguments or input fields, must be either nullable or have a default value.
 
 ## 🗳️ Alternate solutions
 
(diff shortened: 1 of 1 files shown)


rfcs/OptInFeatures.md
```diff
@@ -0,0 +1,152 @@
...
+  name: String
+
+  # [...] other fields omitted for clarity
+
+  # includeOptIn is a list of features to include
+  fields(includeDeprecated: Boolean = false, includeOptIn: [String!]): [__Field!]  
+}
+```
+
+Tools can get a list of `@optIn` features required to use a field (or input field, argument, enum value) using `requiresOptIn`:
```
jturkel, line 83: Do tools also need a mechanism for listing all of the `@optIn` features in a schema? I'm imagining tools like GraphiQL showing users a list of possible features they can opt into rather than requiring users to somehow obtain this information via documentation or some other non-introspection based mechanism.

rfcs/OptInFeatures.md
```diff
@@ -0,0 +1,152 @@
...
+```graphql
+"""
+Indicates that the given field, argument, input field or enum value requires
+giving explicit consent before being used.
+"""
+directive @optIn(feature: String!) repeatable
+    on FIELD_DEFINITION 
+    | ARGUMENT_DEFINITION 
+    | INPUT_FIELD_DEFINITION 
+    | ENUM_VALUE
```
jturkel, line 50: Could `@optIn` also apply to type definitions since you might want to expose entirely new types in an experimental feature?

rfcs/OptInFeatures.md
martinbonnin: Good point 👍 If types are not included in introspection in the first place, there's no way of knowing what possible values are for `includeOptIn`...

That'd be an argument for [marker directives](https://github.com/graphql/graphql-wg/blob/6a47b782a5d88f83dd6004d9d4e9385e2e2ee940/rfcs/OptInFeatures.md#marker-directives) because these user-defined directive would be returned no matter what.

The other way to look at this would be to return all the fields all the time (no more `includeOptIn`) and let the tooling (GraphiQL, Playground, etc...) use the logic they find appropriate to show them in autocomplete (or hide them)? I feel like that'd be my favorite solution. I never really understood why some tools would set `includeDeprecated: false` for an example. Is there a place where there is more context about this?

rfcs/OptInFeatures.md
martinbonnin: Certainly? I copy/pasted this list from `@deprecated` because I started with `@experimental` being the counterpoint of `@deprecated` but as we're moving away from this I think it makes a lot of sense. 

I guess it has implications in terms of validation though. If a type is `@optIn("foo")`, should all the usages be `optIn("foo")` too? That can make a lot of directives?

rfcs/OptInFeatures.md
```diff
@@ -0,0 +1,152 @@
...
+includeOptIn.containsAll(field.requiresOptIn) && (includeDeprecated || !field.isDeprecated)
+```
+
+### Validation
+
+Similarly to [deprecation](https://spec.graphql.org/draft/#sel-FAHnBZNCAACCwDqvK), the `@optIn` directive must not appear on required (non-null without a default) arguments or input object field definitions.
+
+In other words, `@optIn`  arguments or input fields, must be either nullable or have a default value.
+
+## 🗳️ Alternate solutions
```
tomgasson, line 109: ### Alternative proposal: @requiresOptIn and @optIn

Same as the _Proposed solution_, except the schema uses `@requiresOptIn` instead
```graphql
directive @requiresOptIn(feature: String!) on FIELD_DEFINITION 
```
```graphql
type Session {
  id: ID!
  title: String!
  startInstant: Instant @requiresOptIn(feature: "experimentalInstantApi")
  endInstant: Instant @requiresOptIn(feature: "experimentalInstantApi")
}
```
Operations use `@optIn`
```graphql
directive @optIn(feature: String!) on FIELD 
```
```graphql
query {
  session {
    id
    title
    startInstant @optIn(feature: "experimentalInstantApi")
    endInstant @optIn(feature: "experimentalInstantApi")
  }
}
```
Adding `@requiresOptIn` to an existing field would need to be detected as a schema breaking change.
Removing `@requiresOptIn` is safe, and an operation using `@optIn` on a field without `@requiresOptIn` would continue to work.
Tools can use introspection values to provide code actions to add `@optIn` where necessary
Matching the `feature` prevents misuse, for example:
```graphql
query {
  session @optIn {
    id @optIn
    title @optIn
    startInstant @optIn
    endInstant @optIn
  }
}
```

For introspection, unlike the _Proposed solution_ the schema is telling you about the opt-in features, rather than you having to provide them
```diff
 type __Type {
   # [...] other fields omitted for clarity
   fields(
     includeDeprecated: Boolean = false,
+    includeRequiresOptIn: Boolean = false
   ): [__Field!]
 }
 
 type __Field {
   name: String!
   # [...] other fields omitted for clarity
   isDeprecated: Boolean!
   deprecationReason: String
+  requiresOptIn: Boolean!
+  optInFeature: String
 }
```

Pros:
  - Self-documenting. There's no set of feature names defined somewhere else
 - Usage is colocated with consumption
 - No grammar changes
 - No need to enumerate _all opt in features_

Cons:
 - Only uses a string, no capability for providing more type information
 - Goes further than just being documentation to somewhat imply behaviour (although that could still be for server to decide)
  - No place to document a feature itself (i.e one feature might apply to many fields, and want a high-level description of the feature itself)

rfcs/OptInFeatures.md
```diff
@@ -0,0 +1,152 @@
...
+### The `@optIn` directive
+
+It is proposed to add an `@optIn` directive to the specification:
+
+```graphql
+"""
+Indicates that the given field, argument, input field or enum value requires
+giving explicit consent before being used.
+"""
+directive @optIn(feature: String!) repeatable
```
tomgasson, line 46: Should these be be `repeatable`? I'm not sure if that's necessary and it could complicate things

rfcs/OptInFeatures.md
tomgasson: I'm not sure about type definitions.

You can query over a union or interface, which could mean you get one of many types. Specifying what a server would have to do if you hadn't opt-ed in based _on the runtime value_ makes this much more complex. 
I think having a decidable result based only on schema & operation (and regardless of the value) is desirable.

For the same reason, I'm also not sure about `ENUM_VALUE`

ARGUMENT_DEFINITION and INPUT_FIELD_DEFINITION make sense from a static validation perspective, but I'm unsure if there's real-world need for them. 

At the field level it's often things like "this it's hitting a t1.nano and not ready for 1 million users". Experimental input fields and arguments are more likely about business logic and validation correctness. If they are truely doing something newly _experimental_ they could create a new field instead. I'd love to see some examples of situations where this would be a useful addition

rfcs/OptInFeatures.md
martinbonnin: > Alternative proposal: `@requiresOptIn`

Woops, sorry I missed this earlier today, I actually just pushed `@requiresOptIn`, I like this a lot, it plays well with `@optIn` and is more explicit

> Operations use `@optIn`

Do we need to put that in the spec? This has execution implication and different teams might want to implement that differently. 

I initially envisioned the proposal as a way to "communicate" opt-in requirements (like `@deprecated` communicates the deprecated status). "Enforcing" it seems another steps with a wide range of implications.

> For introspection, unlike the Proposed solution the schema is telling you about the opt-in features, rather than you having to provide them

Not sure I get this one. The current proposed solution already exposes the opt-in features:

```graphql
type __Field {
  name: String!
  isDeprecated: Boolean!

  # [...] 

  # list of @requiresOptIn features required to use this field
  requiresOptIn: [String!]
```

We can use an additional Boolean but it feels like duplication because we can tell if there's the need for duplication if the list is not empty. 

On a separate (but related) note, I wish `isDeprecated` would be replaced by checking for `deprecationReason != null`.

rfcs/OptInFeatures.md
martinbonnin: Fair point. Since there are different features I figured out it was possible to require multiple of them and it didn't feel too complex but we can start simple and add `repeatable` down the road if we want to I guess. Although that circles back to the introspection discussion: if we add `repeatable`, then we need to model features as a list which might be breaking, etc...

rfcs/OptInFeatures.md
```diff
@@ -0,0 +1,152 @@
...
+}
+```
+
+Pros:
+* gives more control to the user about the directive used
+* has more type information
+
+Cons:
+* more complex
+* requires a grammar change
```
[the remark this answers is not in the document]
martinbonnin: > Is my understanding correct?

100% correct. I took the liberty of reusing your wording as it's much clearer 👌 and pushed it to the proposal, I hope you don't mind?

> Do clients do something to consume this? Or is it just documenting?

Simply documenting.  See also [my other comment](https://github.com/graphql/graphql-wg/pull/1006#discussion_r889565943). 

> It's not clear what extra information would be useful to capture. Some examples might help clarify

To me, the main advantage is that tools can get the list of opt-in features by scanning the directive definitions. They don't have to scan each field and infer unique feature names. So it's a bit easier on the tooling side.

As a side effect, it also allows users to add arguments to their directives (your `WorkspaceKind` above) although at this point, I'm not sure at all there are use cases for this. 

> This appears to hinder the ability to do out-of-band feature opt-ins (HTTP headers) because you don't have an identifier to use. E.g.

You can always use directive name. You example would actually work

```
headers: {
  "X-ExperimentalApi": "experimentalDeploymentApi"
}
```

benjie:
@martinbonnin Ping me when you want me to merge this. In general RFC documents get merged quickly so that others can raise PRs against them. Also please add yourself as the champion near the top of the document 👍

martinbonnin:
> so that others can raise PRs against them

Ah, that makes a lot of sense 👍 . Feel free to merge for me.

@tomgasson I already included some of your feedbacks there (`Atlassian` prior work and `@requiresOptIn`). Let me know if you want me to revert these commits so that you can get proper attribution.

benjie:
Since @tomgasson thumbs-up'd your comment I'm taking that as a signal to merge 👍

LouisCAD:
For the case of Kotlin, I think you mentioned `@OptIn` in place of `@RequiresOptIn`, which is the replacement for `@Experimental`.

martinbonnin:
@LouisCAD good catch! https://github.com/graphql/graphql-wg/pull/1013 fixes it

PR #1035, opened 2022-07-07
RFC: structs

benjie:
> Symmetric-Transput Replete Unadulterated Composite Type (S.T.R.U.C.T.)
> Spectacular Technology for Representing the Ultimate in Composite Types (S.T.R.U.C.T.)
> Structured, Type-safe, Returnable and Union-Capable Type (S.T.R.U.C.T.)

Structs are essentially expanding the capabilities of the `input` type to output, enabling:

- polymorphic inputs that match outputs
- non-infinite recursion
- safe "wildcard"-style selection over pure data
- "scalars with structure" - a composite type that can be output from GraphQL and input back into GraphQL without modification

For more, please read the RFC.

This is also a precursor to the [struct-based metadata RFC](https://github.com/graphql/graphql-wg/pull/1036).

### DIFF
--- a/rfcs/Struct.md
+++ b/rfcs/Struct.md
@@ -1,0 +1,464 @@
+# Polymorphic-capable composite symmetric input/output type (`struct`)
+
+> _a.k.a "Symmetric-Transput Replete Unadulterated Composite Type" S.T.R.U.C.T._
+
+GraphQL currently has two types that are suitable for both input and output:
+scalar and enum. These are both "leaf" types that generally[1] possess no
+structure.
+
+[1]: Custom scalars may possess internal structure, a common example of this is
+the 'JSON' scalar, however this structure is not defined by nor can it be
+queried via GraphQL (`specifiedBy` notwithstanding).
+
+GraphQL currently has one _composite_ type that is suitable for input: the input
+object type. This type possesses structure, but is only available on input (and
+does not support polymorphism).
+
+This proposal is for a composite type (a type with structure) which is available
+on both input and output, and which supports polymorphism.
+
+## Why?
+
+A polymorphic-capable composite type that's available on both input and output
+would address a number of use cases, so lets dig into them individually:
+
+### Scalar with structure
+
+Sometimes the data that a leaf field resolver returns may have structure,
+represented via something like a custom scalar (e.g. 'JSON'). It would be nice
+to share the schema of this data with the client to enable the client to
+understand it better; this may be solved in part with `@specifiedBy`, but it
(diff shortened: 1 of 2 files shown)


rfcs/Struct.md
```diff
@@ -0,0 +1,464 @@
...
+Struct selection sets are similar to, but not the same as, regular selection
+sets. In particular:
+
+- struct fields are not `FIELD`s, i.e. they cannot have `FIELD` directives
+  attached (suggest we call them `STRUCT_FIELD`)
+- aliases are not allowed, otherwise the simple field merging (see below) does
+  not work (also there's no need for aliasing because struct fields are pure
+  data and do not accept arguments, so they cannot produce multiple values for
+  the same field)
+
```
rivantsov, line 348: that does not seem convincing, to have struct selection field different from regular fields.
No field directives - why not? something like @include? As for aliases, you argue they are not so useful - so what? Forbidding dirs and aliases brings more problems than benefits - users will keep forgetting that cannot use them with structs. Better treat struct fields as regular fields, just with some type restrictions

rfcs/Struct.md
```diff
@@ -0,0 +1,464 @@
...
+difference is intent; because `struct` is intended to be used for input/output
+it's intended that you pull it down, modify it as need be, and then send it back
+again. As such, adding a non-nullable field to a struct used in this way may not
+be a breaking change - existing clients should automatically receive the
+additional fields, and send them back unmodified.
+
+The semantics of `struct` and input objects are very similar, so it could be
+that we repurpose input objects for struct usage. This will need careful
+thought, discussion and investigation. Not least because the term `input` would
+make it confusing ;)
```
rivantsov, line 464: another possibility - making 'struct' a new synonym of 'input', while also allowing input types to be used in outputs. This is non breaking change all the way, and we avoid introducing new concept, only an alias for existing concept.

benjie: Yes, just using the `input` type for structs is one of the potential options. It's not clear if structs actually need to differentiate from input at all. I've used the keyword `struct` for clarity (and to make the text stand even if we decide to introduce struct as a whole new type), but I'm personally on the page of just using input objects.

rivantsov: it's just if input types are allowed in the output, the term 'input' becomes confusing; so we provide 'struct' synonym, recommend it for future use, but never deprecate 'input' for backward compatibility. Struct or something else.. we'll see, but I like struct, it is a good name for the newly re-purposed input type.

leebyron:
Merging to get this draft in place! Feel free to keep the discussion going and add changes

PR #1036, opened 2022-07-07
RFC: struct-based metadata

benjie:
This proposal builds on #1035 to provide GraphQL with a powerful and elegant schema metadata solution.


leebyron:
Merging with policy of liberally merging RFCs so we unblock adding change PRs. Not a signifier of RFC level

Issue #825, opened 2021-12-17
[TSC] RFC: Define a process for reporting and dealing with security vulnerabilities


benjie:
We don't seem to have an official way of dealing with security vulnerability reports currently, nor, as far as I can tell, an official way of reporting them. I believe it should be the TSC's responsibility to deal with security vulnerabilities and alert the relevant people, but as far as I know we don't really have an internal process for this currently. I'd like to propose to the TSC that we address this.

@brianwarner Has proposed the following text (which I have lightly edited) as a starting point:

> "The GraphQL TSC is in an elevated position of trust within the GraphQL community. Security concerns that impact repos under the [`graphql` GitHub org](https://github.com/graphql/) (including reference implementations and official tools) may be responsibly disclosed to the TSC via [any current TSC member](https://github.com/graphql/graphql-wg/blob/main/GraphQL-TSC.md#tsc-members-1), with the expectation that they will be discussed and triaged by the TSC as a whole.
> 
> Our goal is to provide complete, accurate, and actionable disclosures once a reported issue has been sufficiently understood and there has been a reasonable opportunity to deploy fixes responsibly. At no time should a TSC member release information on a pre-disclosed vulnerability to anyone besides other TSC members, Foundation staff, legal counsel, or required authorities unless there is consensus to do so. A TSC member may call for a formal vote to determine an appropriate path forward at any time in the process, if needed.
> 
> In the case of responsible disclosures, the TSC is expected to work in good faith toward a resolution that is in the best interest of the community, including coordinating with maintainers on pre-disclosure patches and the CVE process. As responsible and knowledgable stewards of the GraphQL ecosystem, the TSC is empowered to negotiate the priority level and timelines for announcements and fixes.
> 
> In the case of irresponsible disclosure, regardless of the circumstances, the TSC is expected to make themselves available to convene urgently and to decide upon a communications and action plan."

I plan to present this at the next GraphQL TSC meeting (which is the next GraphQL WG meeting) - I'll be filing the agenda item shortly.

acao:
Thanks for doing this! Let me know how I can help! From this we can create a security policy document add to each of the repos.

saihaj:
related https://github.com/graphql/graphql-js-wg/issues/42

PR #1105, opened 2022-08-29
Updating the decision of the Input Union RFC to RFC #825 using @oneOf

mcwoodle:
## Problem
Through search, I kept coming up with this older Input Union RFC when searching for the natural term of "GraphQL Input Union".

I suspect this will be the most popular search term that those unfamiliar with the WG will try, and @oneOf does not meet that naturally at.  The first search result on google for me is https://github.com/graphql/graphql-spec/issues/488 - which points to this InputUnion file

## This Change

This change just updates the final decision to note the updated RFC for @oneOf which is currently undergoing active dev for inclusion into the release spec.

### DIFF
--- a/rfcs/InputUnion.md
+++ b/rfcs/InputUnion.md
@@ -1124,8 +1124,10 @@ A quick glance at the evaluation results. Remember that passing or failing a spe
 
 # ☑️ Decision Time!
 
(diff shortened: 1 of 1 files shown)


leebyron:
Thanks for this addition!

Have you hit any trouble with the CLA?

benjie:
@oneOf is now merged :tada:

PR #1107, opened 2022-08-30
Add more discussion time for defer/stream

robrichard:


IvanGoncharov:
@robrichard This topic deserves more time, especially since spec PR is now updated.

PR #1125, opened 2022-09-12
Implement new WG times + secondary meetings

leebyron:
This is an implemented final proposal for [Changed Working Group times](https://github.com/graphql/graphql-wg/discussions/1051)

As a short recap, a couple meetings ago the topic was raised to take a fresh look at our meeting cadence based on a few concerns:

- A two hour meeting was a little too long
- Timing was late in the day for EU
- Timing was middle of the night for APAC
- Agendas ran over, without a path for overflow
- Significant progress is being made, and a monthly cadence isn't frequent enough.

This changes timing as follows:

- The existing WG meeting is now named "Primary", remains on first Thursdays, but is reduced to 1.5hr and shifted slightly earlier.
- Two new "Secondary" WG meetings are added, one at the same time on third Thursdays, and one +5hr on second Wednesdays to be more APAC available.

This also revises agenda templates and file system:

- Year/Month/ becomes a directory with named meeting agenda files within
- Updated intro text describing the meeting purpose
- Updated the timeanddate links to use the newer meeting scheduling tool.
- List times in Pacific Time instead of UTC, since that is how our calendar entries have been working.

Finally, this adds tables of all meetings for WG and subcommittees in the repo README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,11 @@
-GraphQL Working Group
-=====================
+# GraphQL Working Group
 
-GraphQL WG (Working Group) is a monthly virtual meeting of maintainers of
-commonly used GraphQL libraries and tools and significant contributors to the
-GraphQL community, operated by the GraphQL Foundation.
+GraphQL WG (Working Group) is a set of recurring virtual meetings of maintainers
+of commonly used GraphQL libraries and tools and significant contributors to the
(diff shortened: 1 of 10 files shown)


benjie (changes requested):

benjie:
LGTM! Let's add GraphiQL and then we should be good to go :raised_hands:

PR #1141, opened 2022-10-12
add dates to wg agenda file names

yaacovCR:
motivation: we can now determine the date of the meeting simply from the path + filename of the agenda file, similar to when there was a more sparse wg cadence


benjie:
I suspect the reason for not doing this is that the secondary meeting is on two different dates for its target audience - Wednesday for US and Thursday for APAC.

yaacovCR:
Closing until a better solution lightbulbs. :)

PR #1155, opened 2022-10-21
Add user-content use-case to Struct RFC

Kingdutch:


rfcs/Struct.md
```diff
@@ -531,4 +531,77 @@ reasonably prefer to write a query like this:
...
+
+The schema for this example would be similar to the schema provided in the Struct
+Union excample. The use-case does highlight one drawback with respect to the
+constraints outlined in the use-case.
+
+Since `struct`s can not themselves contain `object`s providing the structured
+content in this manner would mean that the client could not request optimised
+media (one of the reasons why a pre-rendered output was undesirable in the first
+place), or the image type itself must be a struct which seems to disallow
+arguments fields which is a common pattern to request specific transofmrations of
```
benjie, line 568:
```suggestion
arguments fields which is a common pattern to request specific transformations of
```

benjie (changes requested):

benjie (approved):
Thanks for the use case!

PR #1167, opened 2022-11-03
November notes

benjie:
Here's the notes from the primary meeting in November; the secondary meetings notes should be merged in when ready.


benjie:
cc @leebyron Should we have one notes file for all three meetings to mirror the GDocs situation?

PR #1173, opened 2022-11-10
Remove references agenda item

yaacovCR:
Reference are perhaps most useful for an alternative implementation of incremental delivery that does not branch execution, which I am actively working on.

I hope to reintroduce references perhaps retooled, when that work is ready.


benjie:
We did mention this topic at the meeting when reviewing the agenda, so I'd rather leave it as a historical record. Is that okay?

yaacovCR:
Of course!

PR #1198, opened 2022-12-01
Move agenda and notes into annual directories

leebyron:
Curious for feedback. This moves historical agenda and notes into annualized directories. This cleans up the directory structure quite a bit however may also lead to broken links (this does not attempt to find and fix those).


mjmahone:
I am a fan of this change. I get breaking links is always annoying, but I'm not sure there *are* a lot of links floating around?

An alternative would be to copy the old files into directories and leave the existing files around for a while. Or make them symlinks. It's not a huge issue to have a bunch of files in the root directory, as directories are usually sorted to the top, but I don't have strong enough opinions to say "yes definitely symlink/copy because..." or "no definitely don't because..."

mjmahone (approved):

PR #1248, opened 2023-01-31
Add status check for default value PR stack

yaacovCR:
should just take a minute or two, just want to re-iterate the call(s?) to action involved

### DIFF
--- a/agendas/2023/02-Feb/02-wg-primary.md
+++ b/agendas/2023/02-Feb/02-wg-primary.md
@@ -104,6 +104,7 @@ hold additional secondary meetings later in the month.
 | Ivan Goncharov   | @IvanGoncharov | Apollo            | Lviv, UA              |
 | Matt Mahoney     | @mjmahone      | Meta              | New York, NY, US      |
 | Rob Richard      | @robrichard    | 1stDibs           | Jersey City, NJ, US   |
+| Yaacov Rydzinski | @yaacovCR      | Individual        | Neve Daniel, IL       |
 
 ## Agenda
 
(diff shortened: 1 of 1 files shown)


agendas/2023/02-Feb/02-wg-primary.md
```diff
@@ -135,3 +136,8 @@ hold additional secondary meetings later in the month.
 1. `@defer`: fully deduplicate response (30m, Ivan)
    - [RFC] (https://github.com/robrichard/defer-stream-wg/discussions/65#discussioncomment-4757697)
    - GraphQL.js implementation: TBD
+1. Default Value Validation Status Check (5m, Yaacov)
+   - [Lee's original PR stack on graphql-js](https://github.com/graphql/graphql-js/pull/3049)
+   - [Now rebased on main](https://github.com/graphql/graphql-js/pull/3814)
+   - [Benjie's original PR for graphql-spec](https://github.com/graphql/graphql-js/pull/3049)
+   - [Now rebased on main](https://github.com/graphql/graphql-spec/pull/793)
```
benjie, line 143: These links appear to be pointing to the wrong places.

I've also just merged `main` into my branch so your rebase isn't needed - best to keep the discussion in one place.

benjie (changes requested):

agendas/2023/02-Feb/02-wg-primary.md
yaacovCR: Done!

And I closed https://github.com/graphql/graphql-spec/pull/1012

benjie (approved):

benjie:
@yaacovCR How well do you feel the Spec PR matches up with the GraphQL.js implementation? IIRC (and this was over a year ago, so my memory is very fuzzy on the subject) the GraphQL.js implementation evolved and I wasn't sure whether or not the spec reflected it any more - given you recently rebased both maybe you have a better feel for it?

yaacovCR:
@benjie -- considering the number of changes, it is certainly tough to say, and so the more eyes on board the better!

The main breaking change within graphql-js (besides function names/signatures) is that default values when supplied programmatically should be in the same domain as variable values, i.e. they should be external, and not internal.

To quote @leebyron see https://github.com/graphql/graphql-js/pull/3814 referencing the earlier PR:

>Changes default values from being represented as an assumed-coerced "internal input value" to a pre-coerced "external input value" (See chart below).
> This allows programmatically provided default values to be represented in the same domain as values sent to the service via variable values, and allows it to have well defined methods for both transforming into a printed GraphQL literal string for introspection / schema printing (via valueToLiteral()) or coercing into an "internal input value" for use at runtime (via coerceInputValue())

This seems to still match the spec PR change pretty well!

PR #1252, opened 2023-02-01
Refine TSC voting process

leebyron:
This updates our TSC charter to update and clarify how voting works:

- Updates voting processes to Condorcet Beatpath and STV Meek, fixing an inaccuracy where we described Condorcet as for multiple winners where we meant multiple candidates.
- Adds a table clarifying the possible vote scenarios and where they count for clarity

### DIFF
--- a/GraphQL-TSC.md
+++ b/GraphQL-TSC.md
@@ -173,8 +173,15 @@ After a TSC seat vacancy the remaining TSC will collect self-nominations and vot
 
 When something cannot be decided by consensus or a formal vote is required, the TSC will use a method appropriate to the situation:
 
-- Simple majority for single-winner votes.
-- Condorcet method for multiple-winner votes.
+| Situation                      | Response         | Method                                   |
+| ------------------------------ | ---------------- | ---------------------------------------- |
+| Binary choice                  | Single choice    | Simple majority as per [procedure][]     |
+| Many choices, single winner    | Rank all choices | [Condorcet Beatpath/Schulze][]           |
+| Many choices, multiple winners | Rank all choices | [Single Transferable Vote (STV), Meek][] |
+
+[procedure]: #voting-procedure
+[condorcet beatpath/schulze]: https://en.wikipedia.org/wiki/Schulze_method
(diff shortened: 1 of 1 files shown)


mjmahone (approved):

GraphQL-TSC.md
```diff
@@ -210,6 +217,20 @@ TSC members are not required to vote. There are three ways an _attending_ member
...
+
+| Vote                            | Counts as _attending_ | Counts towards _quorum_ | Counts towards a choice |
+| ------------------------------- | :-------------------: | :---------------------: | :---------------------: |
+| Attending member choice         |          ✅           |           ✅            |           ✅            |
+| Attending member present        |          ✅           |           ✅            |           ❌            |
+| Attending member absent         |          ✅           |           ❌            |           ❌            |
+| Attending member abstain        |          ✅           |           ❌            |           ❌            |
+| Attending member recused        |          ❌           |           ❌            |           ❌            |
+| "Non-attending member" choice   |          ❌           |           ❌            |           ✅            |
+| "Non-attending member" non-vote |          ❌           |           ❌            |           ❌            |
```
benjie, line 232: It took me a moment to digest this, so here's a brief summary of my understanding:

"Attending member" and "non-attending member" refer to whether the TSC member has attended recent WG meetings or not (doesn't relate to attendance at the vote).

For the vote you can have one of 4 statuses (the default being "absent"):

- choice - you say "yes" / "no" (or whatever is appropriate for the vote) and your vote is counted
- present - you say "present" and thus you count for reaching quorum but not as a vote
  - helps the vote progress
- absent/abstain (equivalent) - you don't count towards quorum nor towards the vote
  - If you don't want to vote for/against something but you want the vote to conclude then you should say "present" instead so you count towards quorum
- recused - if you have a conflict of interest then you should recuse yourself and we all pretend you don't and never have existed for the purposes of the vote/quorum/etc 😉 

More about present/absent/abstain/recused here: https://github.com/graphql/graphql-wg/blob/main/GraphQL-TSC.md#non-votes

PR #1355, opened 2023-07-20
Update subcommittees

benjie:
Incremental Delivery WG has been meeting for a while now. CCN has just started meeting (time may change). Two subcommittees are on hiatus as no-one has brought any agenda items to them for a while now, so I've moved them to a new section.


fotoetienne (approved):
👍 Thanks benjie!

PR #1358, opened 2023-07-31
RFC: Trim down CCN proposal

calvincestari:
This trims down the Client-Controlled Nullability proposal back to the most basic form. This is a result of recent CCN sub-WG meetings trying to progress the proposal through the stages.

More detail to come in the August 3, 2023 WG meeting.

### DIFF
--- a/rfcs/ClientControlledNullability.md
+++ b/rfcs/ClientControlledNullability.md
@@ -264,6 +264,23 @@ precedent.
 
 ## Decision Log
 
+### `!` as the only designator in the RFC (July 2023)
+
+At the July 26th meeting of the CCN sub-WG it was decided to simplify the RFC for the sake of progress. This 
+reverses the "`?` as a counterpart to the `!`" designator.
+
+The increased scope of the RFC with the nullability designator `?` has stalled the RFC for more than two 
+years ([the initial RFC from April 2021](https://github.com/twof/graphql-spec/pull/1/files)).
+
+One of the contentious discussions we encountered was around the proposed behaviors of the nullability designator 
+`?`. Regardless of the chosen path, it had unintuitive semantics (see graphql/client-controlled-nullability-wg#2), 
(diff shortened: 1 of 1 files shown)


calvincestari:
@twof, I think this captures the changes we spoke about for the proposal body.

I'm not sure what to do for the decision log though; I want to leave the decision log items there to note what was decided in the past but leaving them there could detract from our goal of going back to basics.. Any ideas?

twof:
@calvincestari I would say keep the decision log append-only, adding the most recent decisions to the top, and date the new ones. It's likely we return to some of the features we're scrapping, so having a summary of the decisions we made last time will help with future discussions.

twof (approved):

benjie:
Mark this as ready and ping me when it's ready for merge. RFCs get merged easily, they don't need much review process.

aprilrd (approved):
Changes look good to me too.

> @twof, I think this captures the changes we spoke about for the proposal body.
> 
> I'm not sure what to do for the decision log though; I want to leave the decision log items there to note what was decided in the past but leaving them there could detract from our goal of going back to basics.. Any ideas?

Talking about a decision log, would you like to add a new decision? Something akin to:

### `!` as the only designator in the RFC

CCN WG decided to simplify the RFC at the 2023-07-26 meeting for the sake of progress. This reverses the "`?` as a counterpart to the `!`" decision.

The increased scope of the RFC with the nullability designator `?` has stalled the RFC for more than two years ([the initial RFC from April 2021](https://github.com/twof/graphql-spec/pull/1/files)).

One of the thorny decision points we encountered was around the proposed behaviors of the nullability designator `?`. Regardless of the chosen path, it had unintuitive semantics (see https://github.com/graphql/client-controlled-nullability-wg/issues/2), which require other RFCs, such as [fragment boundaries](https://github.com/graphql/graphql-wg/blob/main/rfcs/FragmentModularity.md), or new ideas like inline errors. 

Even though we recognize the value of having a counterpart to `!`, We made a tradeoff decision to provide value to the community sooner than later.

calvincestari:
@benjie, this should now be ready for merging - thanks.

PR #1370, opened 2023-08-19
Feature Discovery RFC

aprilrd:
I created a strawman feature discovery RFC to aid the discussion at the August CCN WG meeting.

The need for feature discovery was called out at [the August GraphQL WG meeting](https://github.com/graphql/graphql-wg/blob/main/notes/2023/2023-08.md#ccn).


benjie:
Thanks for raising this! In addition to introspection additions, I think there needs to be an SDL indication too. I'm going to go ahead and merge this, any edits can be applied via follow-up PRs.

PR #1380, opened 2023-09-06
Updating CA to BC for the canadian province

itamark:


benjie:
That’s meant to be the country code really 😇

itamark:
@benjie I snuck it by you

PR #1390, opened 2023-09-07
Adding slides from today

itamark:


benjie:
The access to these slides is not public. You might consider File -> Share -> Publish to Web, or alternatively the traditional GDoc sharing with read-only access for anyone with the URL.

Also, please add the link to the notes too: https://github.com/graphql/graphql-wg/blob/main/notes/2023/2023-09.md

itamark:
> The access to these slides is not public. You might consider File -> Share -> Publish to Web, or alternatively the traditional GDoc sharing with read-only access for anyone with the URL.
> 
> Also, please add the link to the notes too: https://github.com/graphql/graphql-wg/blob/main/notes/2023/2023-09.md

Thanks @benjie - made a public copy and updated this PR.

PR #1515, opened 2024-04-11
Update GraphQL-TSC.md to detail voting process

leebyron:
Formalize vote phases, adding more detail to the call and conclude phases, explicitly calling out contact methods and reporting back results.

@graphql/tsc

### DIFF
--- a/GraphQL-TSC.md
+++ b/GraphQL-TSC.md
@@ -200,15 +200,33 @@ Note: A member may be recused (i.e. for a member election) in which case they do
 
 ### Voting process
 
-Because we work in a distributed environment, the voting process must account for a range of time zones and schedules. Once the threshold of a quorum has been met and a vote is valid, one of these two critera must be satisfied to conclude a vote:
+The GraphQL TSC is a distributed group and our voting process must account for a range of time zones and schedules. TSC votes are held asynchronously and follow three distinct phases:
+
+1. Call for vote
+2. Open voting
+3. Conclusion of vote
+
+**1. Call for vote**
+
+- Only a member of the TSC may call for a vote (though any member of the community may join a GraphQL Working Group meeting to propose one)
+- A Github issue should be opened in the appropriate repository detailing the vote along with any relevant context and guidance to voting TSC members. Most votes are public and are held in the [graphql-wg](https://github.com/graphql/graphql-wg) repository. In less frequent cases a private vote may be held in a specific private repository (for example, to address security concerns).
+- A reasonable effort must be made to contact all voting TSC members by the individual calling for a vote. Best practice is to **a)** mention `@graphql/tsc` on the Github issue, and **b)** email all voting TSC members, and then optionally **c)** direct message individuals on preferred messaging platforms.
+
+**2. Open voting**
+
+Once the Github issue requiring a vote is posted and all voting TSC members have been contacted, the vote is open. Individual votes may be publicly or privately cast. Once the threshold of a quorum has been met and a vote is valid, one of these two critera must be satisfied to conclude a vote:
 
 - A notice is sent via email that the vote will conclude in three business days, reminding those who haven't voted that they should do so. The vote will conclude at the end of this time.
 - The election results would not change if all remaining members were to vote.
+
+**3. Conclusion of vote**
 
 Once a valid vote is concluded, the result is determined by the number of votes received at that time (as opposed to the total number of TSC members):
 
 - For a binary choice, the votes in favor must exceed half for a majority (or 2/3 for a supermajority) of the total number of votes.
 - For ranked choices, all votes received at the time the vote is concluded are considered.
+
+The result of the vote must be published back to the original Github issue which called for the vote.
 
 ### Non-votes
 


benjie (changes requested):
I approve of these changes with this minor quibble.

GraphQL-TSC.md
```diff
@@ -200,16 +200,34 @@ Note: A member may be recused (i.e. for a member election) in which case they do
...
+
+1. Call for vote
+2. Open voting
+3. Conclusion of vote
+
+**1. Call for vote**
+
+- Only a member of the TSC may call for a vote (though any member of the community may join a GraphQL Working Group meeting to propose one)
+- A Github issue should be opened in the appropriate repository detailing the vote along with any relevant context and guidance to voting TSC members. Most votes are public and are held in the [graphql-wg](https://github.com/graphql/graphql-wg) repository. In less frequent cases a private vote may be held in a specific private repository (for example, to address security concerns).
+- A reasonable effort must be made to contact all voting TSC members by the individual calling for a vote. Best practice is to **a)** mention `@graphql/tsc` on the Github issue, and **b)** email all voting TSC members, and then optionally **c)** direct message individuals on preferred messaging platforms.
```
IvanGoncharov, line 213: I think we are missing the major part of the algorithm, which is time to present your case or even form your position on the topic. 

For example, I have concerns about some topics but need to formulate them (which takes time), but I go to the issue and see that it is supported by the majority of the votes.
This situation is very discouraging because, at this point, you are going against an already approved decision without knowing if your arguments could influence votes if the people read them before voting.

In all voting systems, I saw, there is a mandatory discussion period before voting.
Recently, I was in a situation where I disagreed with an already-voted decision and raised my concern because I feel strongly against it in its current form.

Imagine the same happens with our current procedure: I go to an issue and see it already has the majority of votes in favor. If I just write my concerns, I am not sure that members who already voted will see them at all (depending on their GitHub settings and willingness to regularly check them), so I was forced to write separate emails to everyone.

This creates an effect where a decision can "rubber stump" very fast, and TSC members that have concerns need to decide if voicing them is worth the effort or not.

The same goes for public votes; if we add something to the agenda on the day of the WG call, only people who joined the WG call for some other reasons would have an option to voice their concerns.

Taking into that currently we vote a few times per year, I don't see any harm in making a rule that stuff can be voted on only if it were added to the agenda (or after it is posted in this repo and TSC members are notified) at least one week before the vote.

GraphQL-TSC.md
```diff
@@ -200,16 +200,34 @@ Note: A member may be recused (i.e. for a member election) in which case they do
...
+The GraphQL TSC is a distributed group and our voting process must account for a range of time zones and schedules. TSC votes are held asynchronously and follow three distinct phases:
+
+1. Call for vote
+2. Open voting
+3. Conclusion of vote
+
+**1. Call for vote**
+
+- Only a member of the TSC may call for a vote (though any member of the community may join a GraphQL Working Group meeting to propose one)
+- A Github issue should be opened in the appropriate repository detailing the vote along with any relevant context and guidance to voting TSC members. Most votes are public and are held in the [graphql-wg](https://github.com/graphql/graphql-wg) repository. In less frequent cases a private vote may be held in a specific private repository (for example, to address security concerns).
```
IvanGoncharov, line 212: We need stronger criteria here, something that prevents situations similar to what happened.
I should be something like:
Vote can be held in private only if there are a significant risk to GraphQL foundation or broader GraphQL ecosystem. It is MANDATORY that desigion record include explanation of potential risk of making this vote public.

For example, in case of security issue it can be as short as:
> This security issue is not disclosed to the public so we can't disclose information about it.

So we need to embrace the "public by default" policy that was advertised, and for that, TSC members suggesting "private vote" should provide explicit arguments for making it.

GraphQL-TSC.md
```diff
@@ -200,16 +200,34 @@ Note: A member may be recused (i.e. for a member election) in which case they do
...
+- A reasonable effort must be made to contact all voting TSC members by the individual calling for a vote. Best practice is to **a)** mention `@graphql/tsc` on the Github issue, and **b)** email all voting TSC members, and then optionally **c)** direct message individuals on preferred messaging platforms.
+
+**2. Open voting**
+
+Once the Github issue requiring a vote is posted and all voting TSC members have been contacted, the vote is open. Individual votes may be publicly or privately cast. Once the threshold of a quorum has been met and a vote is valid, one of these two critera must be satisfied to conclude a vote:
```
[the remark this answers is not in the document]
mjmahone: I'd argue there probably should *not* be privately cast votes. I'd be tempted to say all public votes we should be OK with publicly recording our votes for. For the same reasons you probably don't want your parliamentarian casting private votes.

GraphQL-TSC.md
```diff
@@ -200,16 +200,34 @@ Note: A member may be recused (i.e. for a member election) in which case they do
 
 ### Voting process
 
-Because we work in a distributed environment, the voting process must account for a range of time zones and schedules. Once the threshold of a quorum has been met and a vote is valid, one of these two critera must be satisfied to conclude a vote:
+The GraphQL TSC is a distributed group and our voting process must account for a range of time zones and schedules. TSC votes are held asynchronously and follow three distinct phases:
+
+1. Call for vote
```
leebyron, line 205: Capturing a note from @benjie - need to capture what constitute a vote, and what appears similar but is different (for example, an opinion poll or asking for feedback)

GraphQL-TSC.md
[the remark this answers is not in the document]
benjie: Whilst I agree with you in general, there are times when votes being public may unfairly sway the vote - for example if the vote relates to a person, project or organization then personal relationships may influence a public vote whereas a private vote, hopefully, can be cast without fear of reprisal. That said, deciding up front whether something is a public or private vote (for all participants) and allowing TSC members to recuse themselves seems reasonable.

GraphQL-TSC.md
martinbonnin: +1 to Ivan's point of adding the item to a wg agenda one week before the vote.

I would also remove the mention of emails and/or DMs. TSC member should be expected to monitor the graphql-wg repo.

benjie: +1 to notice period 

-1 to reducing notification of TSC members that a vote is happening and hoping they manage to constantly stay up to date with their GitHub notifications (see private discussion from 2023 in https://github.com/graphql/tsc/discussions/33)

martinbonnin: > -1 to reducing notification of TSC members that a vote is happening and hoping they manage to constantly stay up to date with their GitHub notifications 

Sounds good to me 👍

PR #1593, opened 2024-11-22
Glossary: add 'implementation' 

martinbonnin:
Adding implementation to the Glossary. Feedbacks welcome!

### DIFF
--- a/rfcs/glossary/Appendix C -- Glossary.md
+++ b/rfcs/glossary/Appendix C -- Glossary.md
@@ -201,8 +201,8 @@ can happen anywhere as long as it conforms to the specification.
 
 ### (GraphQL) implementation
 
-**Definition**: a software that implements the algorithms and behaviour described
-in the specification. 
+**Definition**: software that implements the algorithms and behaviours described
+in the GraphQL specification. 
 
 Examples: graphql-js, Hot Chocolate, graphql-java, etc...
 
--- a/rfcs/glossary/Appendix C -- Glossary.md
+++ b/rfcs/glossary/Appendix C -- Glossary.md
@@ -187,8 +187,7 @@ Example:
 
 ### (GraphQL) service
 
-**Definition**: anything capable of executing a GraphQL request and returning
-a GraphQL response conforming to the specification.
+**Definition**: A GraphQL service generates a [response](https://spec.graphql.org/draft/#sec-Response) from a [request](https://spec.graphql.org/draft/#request) via [execution](https://spec.graphql.org/draft/#sec-Execution).
 
 How the conformance is implemented is irrelevant. A cache in front of another
 GraphQL service can itself be seen as a conforming service, assuming it does not
--- a/rfcs/glossary/Appendix C -- Glossary.md
+++ b/rfcs/glossary/Appendix C -- Glossary.md
@@ -201,5 +201,10 @@ can happen anywhere as long as it conforms to the specification.
 **Definition**: the software used to implement a service. Implementations may be 
 schema-first or code-first.
 
+An implementation "implements" the specification if the services it produces 
+conforms to the GraphQL specification. An implementation may implement the 
+specification but not some of its optional features. Some implementations 
+implement the specification including all optional features.
+
 Examples: graphql-js, Hot Chocolate, graphql-java, etc...
 
--- a/rfcs/glossary/Appendix C -- Glossary.md
+++ b/rfcs/glossary/Appendix C -- Glossary.md
@@ -190,8 +190,9 @@ Example:
 **Definition**: anything capable of executing a GraphQL request and returning
 a GraphQL response conforming to the specification.
 
-How the conformance is enforced is irrelevant. An HTTP cache can be a conforming
-service.
+How the conformance is implemented is irrelevant. A cache in front of another
+GraphQL service can itself be seen as a conforming service, assuming it does not
+break any conformance requirements (e.g. on cache failure).
 
 Examples: GitHub GraphQL API, Shopify GraphQL API, etc... 
 
--- a/rfcs/glossary/Appendix C -- Glossary.md
+++ b/rfcs/glossary/Appendix C -- Glossary.md
@@ -187,10 +187,13 @@ Example:
 
 ### (GraphQL) service
 
-**Definition**: a software capable of executing a GraphQL request and returning
-a GraphQL response according to this specification. 
+**Definition**: anything capable of executing a GraphQL request and returning
+a GraphQL response conforming to the specification.
 
-Examples: GitHub GraphQL API, Shopify GraphQL API, etc...
+How the conformance is enforced is irrelevant. An HTTP cache can be a conforming
+service.
+
+Examples: GitHub GraphQL API, Shopify GraphQL API, etc... 
 
 Note: The specification has no concept of client and server. While most 
 services are implemented as servers, this is not a requirement and execution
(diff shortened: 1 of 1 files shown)


benjie (changes requested):
:thinking: Does an implementation have to be used to build a service? `graphql-js` implements the parsing section of the GraphQL spec, and can be used for building developer tooling, without ever being used to build a GraphQL service. I don't think this feels quite right yet...

rfcs/glossary/Appendix C -- Glossary.md
```diff
@@ -187,9 +187,22 @@ Example:
 
 ### (GraphQL) service
 
-**Definition**: a software capable of executing a GraphQL request and returning
-a GraphQL response according to this specification. 
+**Definition**: anything capable of executing a GraphQL request and returning
+a GraphQL response conforming to the specification.
+
+How the conformance is enforced is irrelevant. An HTTP cache can be a conforming
+service.
```
benjie, line 194:
```suggestion
How the conformance is implemented is irrelevant. A cache in front of another
GraphQL service can itself be seen as a conforming service, assuming it does not
break any conformance requirements (e.g. on cache failure).
```

rfcs/glossary/Appendix C -- Glossary.md
```diff
@@ -187,9 +187,22 @@ Example:
...
+Examples: GitHub GraphQL API, Shopify GraphQL API, etc... 
 
 Note: The specification has no concept of client and server. While most 
 services are implemented as servers, this is not a requirement and execution
 can happen anywhere as long as it conforms to the specification.
+
+### (GraphQL) implementation
+
+**Definition**: a software that implements the algorithms and behaviour described
+in the specification.
```
benjie, line 205:
```suggestion
**Definition**: software that implements the algorithms and behaviours described
in the GraphQL specification. 
```

benjie (changes requested):
Maybe this captures the intent more precisely?

rfcs/glossary/Appendix C -- Glossary.md
martinbonnin: More details is always good, thanks!

martinbonnin:
I'm a slow writer but I wanted to say this:

I simplified and left "optional features" out on purpose for now cause I'm still not 100% clear about it. 

```
GraphQL implementations that support the type system definition language should provide the @specifiedBy directive if representing custom scalar definitions.
```

I'd love the spec to be more detailed about what an implementation is and what "support the type system definition language" means in this context. Is this for schema-first libraries? If yes, what should code-first libraries do? Or is this something else? What does "provide" mean? Expose them in introspection? Allow validation of the incoming SDL? Or maybe both?

benjie:
> I'd love the spec to be more detailed about what an implementation is and what "support the type system definition language" means in this context. 

I'd say an implementation is anything that implements the GraphQL spec: a service, some tooling, etc

> Is this for schema-first libraries?

Yes, but for all other SDL-writing use cases; client-side schema extensions; tools that take introspection data and represent it as SDL (e.g. so it can be used by tooling such as graphql-schema-generator); services that allow extensions via SDL even if the original service is not SDL-first; etc.

> If yes, what should code-first libraries do?

In GraphQL.js you'd do `new GraphQLScalarType({name: "...", specifiedByURL: "...", ...})` (note it does not use a directive because it's not SDL).

> What does "provide" mean?

It means make it so that the directive can be used; e.g. for this document to be valid:

```graphql
scalar MyScalar @specifiedBy(url: "...")
```

the implementation must provide the `@specifiedBy` directive. Otherwise you'd get an unknown directive error.

calvincestari:
If we go back to the beginning - why is a distinction necessary? How does it change interpretation or conformance rules of the spec?

martinbonnin:
> If we go back to the beginning - why is a distinction necessary? How does it change interpretation or conformance rules of the spec?

Wondering about this too. If we're saying "service" == "implementation" everything would be much simpler.

The spec has language specific to implementations though:

``` 
GraphQL implementations that support the type system definition language should provide
the @specifiedBy directive if representing custom scalar definitions.
``` 

That doesn't make any sense to someone that just wants to have a conforming server. 

IMO the spec adresses 2 use different cases:

**Service** 

```js
function executeRequest(request: GraphQLRequest): GraphQLResponse { ... }
```

**Implementation**
```js
function buildSchema(sdl: String, resolvers: Resolvers) GraphQLSchema { ... }
// potentially more (parseAst, validateOperation, etc...)
``` 

A `GraphQLSchema` should have an `execute()` function so is technically a service.
I'd say an "implementation" may be thought as a super set of a "service"...

The important thing is that we don't want service to have to conform to all the "implementation" language. At least this is my current understanding of it but it's way different from what it was yesterday so it might still change 😅 ...

calvincestari:
> The spec has language specific to implementations though

Is this intentional though, or has it simply been used as the spec evolved without a common language?

rfcs/glossary/Appendix C -- Glossary.md
```diff
@@ -187,9 +187,23 @@ Example:
 
 ### (GraphQL) service
 
-**Definition**: a software capable of executing a GraphQL request and returning
-a GraphQL response according to this specification. 
+**Definition**: anything capable of executing a GraphQL request and returning
+a GraphQL response conforming to the specification.
```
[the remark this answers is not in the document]
martinbonnin: Perfect 👌

martinbonnin:
> all services are implementations

I can lean into that but I'd say it's not the intuitive mental model. My personal anecdotal mental models is "the GitHub API is a service", "graphql-js is an implementation". The spec also mentions "service implementation" a couple of times which hints at different things.

> an implementation may support fewer (or different) GraphQL features than a service

I never realized that. TBH, I always thought of an implementation as something that has the same feature set as `graphql-js`. As a user, I'd be quite suprised to come into a partial GraphQL implementation. 

From your last comment, I understand a simple GraphQL lexer is technically a compliant GraphQL implementation? I would expect an implementation to support at least parsing + validation + execution to claim the "implementation" status.

> What language in the GraphQL spec do we want a service to not have to conform to?

The bone I'm currently chewing on is [here](https://spec.graphql.org/draft/#sel-EAHnBPDFAADHC9Bnwe):

``` 
GraphQL implementations should provide the @skip and @include directives.

GraphQL implementations that support the type system definition language must provide the @deprecated directive if representing deprecated portions of the schema.

[...]

``` 

This can't be for services since services don't care about SDL.

calvincestari:
> TBH, I always thought of an implementation as something that has the same feature set as `graphql-js`.

As the reference implementation `graphql-js` is not representative of any implementation that can claim to be spec compliant. In `graphql-js`-world there is no _should_ or _may_ in the spec, everything is _must_. 

> As a user, I'd be quite suprised to come into a partial GraphQL implementation.

A partial implementation is only one that ignores the _must_ (required) features no?

martinbonnin:
> As the reference implementation graphql-js is not representative of any implementation that can claim to be spec compliant. 

This is what I'm questioning. I think from a wider community point of view, it's a reasonnable expectation that `implementation == graphql-js like`. graphql-java is very similar to graphql-js. I want to say Hot Chocolate too, etc... If I want to do GraphQL in language X, I expect a library that has the same features as the reference implementation, it's a relatively established pattern. 

If now we're saying a GraphQL client or lexer is a GraphQL implementation, it might be the technical appropriate spec term but also a confusing one for the newcomer.

martinbonnin:
Looks like there's no consensus here. Closing this one and I'll reopen a new one if needed.

PR #1597, opened 2024-11-30
Glossary: add "error" and "field error"

martinbonnin:
Add `error` and `field error` to the Glossary.

I contemplated the possibility to include `internal error` as well but TBH I'm not sure I understand the nuance between `error` and `internal error`.

### DIFF
--- a/rfcs/glossary/Appendix C -- Glossary.md
+++ b/rfcs/glossary/Appendix C -- Glossary.md
@@ -178,7 +178,7 @@ accessing this HTTP endpoint _implicitly_ defines the GraphQL schema to use.
 
 Errors are raised during the execution of an algorithm. Callers of the algorithm may handle errors and recover from the error.
 
-Note: errors can be thought as exceptions in languages that support them.
+Note: Errors can be thought as exceptions in languages that support them.
 
 ### (GraphQL) field error
 
--- a/rfcs/glossary/Appendix C -- Glossary.md
+++ b/rfcs/glossary/Appendix C -- Glossary.md
@@ -172,6 +172,8 @@ accessing this HTTP endpoint _implicitly_ defines the GraphQL schema to use.
 > When using GraphQL over HTTP, it's common to encode the GraphQL request as
 > JSON.
 
+For more details, refer to [the specification definition](https://spec.graphql.org/draft/#request).
+
 ### error
 
 **Definition**: An error is an anomalous condition that breaks the execution of an algorithm.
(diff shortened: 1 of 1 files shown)


rfcs/glossary/Appendix C -- Glossary.md
```diff
@@ -172,6 +172,18 @@ accessing this HTTP endpoint _implicitly_ defines the GraphQL schema to use.
 > When using GraphQL over HTTP, it's common to encode the GraphQL request as
 > JSON.
 
+### error
+
+**Definition**: An error is an anomalous condition that breaks the execution of an algorithm.
+
+Errors are raised during the execution of an algorithm. Callers of the algorithm may handle errors and recover from the error.
+
+Note: errors can be thought as exceptions in languages that support them.
```
benjie, line 181:
```suggestion
Note: Errors can be thought as exceptions in languages that support them.
```

rfcs/glossary/Appendix C -- Glossary.md
```diff
@@ -172,6 +172,18 @@ accessing this HTTP endpoint _implicitly_ defines the GraphQL schema to use.
...
+
+### (GraphQL) field error
+
+**Definition**: A field error is an error raised during the execution of a particular field which results in partial response data. This may occur due to an internal error during value resolution or failure to coerce the resulting value.
```
benjie, line 185: When the spec itself already includes an explicit hyperlinked definition for a term, is it necessary to duplicate it in the appendix manually?

I think many of the definitions in this appendix can be (and have been) migrated to the spec via `:: ` definitions. Perhaps working on a spec-md feature that lists all the definitions as part of the output would be a good idea.

martinbonnin: > I think many of the definitions in this appendix can be (and have been) migrated to the spec via :: definitions. 

100% agree. No need to manually maintain 2 source of truth. I forgot we had the `::` definitions.

I still see the Glossary as a good "incubating" place to discuss the terminology and allow PRs to be discussed/merged more easily while they make their way to something more official.

martinbonnin: I have added a few links to redirect to the official spec when possible. Turns out, a lot of this glossary is also present in the spec but the glossary includes more information. In "request" for an example, there is mention of HTTP and I think it helps the overall understanding.

benjie (approved):

PR #1604, opened 2024-12-05
Add annual agenda items for TSC elections

benjie:
`wgutils` now supports `annualItems`. I've added some items to help us adhere to the TSC election process. We could also add items to remind us to cut spec releases/etc if we wanted to do that on a schedule.

@leebyron the form at https://tsc-nomination.graphql.org/ needs updating each year (including now!) - or make it evergreen?

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   },
   "devDependencies": {
     "prettier": "^2.1.2",
-    "wgutils": "^1.0.0"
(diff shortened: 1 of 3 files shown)


benjie:
I've not linked a GitHub issue about the elections because we'd need to update this each year (and the config should be evergreen); but here's a previous one:

https://github.com/graphql/graphql-wg/issues/765

PR #1607, opened 2024-12-05
Enable agenda automerge workflow

benjie:
See also:

https://github.com/graphql/wgutils/commit/7280d940f1d300605b9e73b2db6e0588274830cc

Essentially this checks the following:

- Required CI checks (i.e. CLAbot) have passed
- The PR only modifies 1 file
- The file is an agenda file
- The PR only adds lines to the file (no deletions)
- The text of the lines added look like they're reasonable characters

Assuming all this passes, it then merges the PR automatically.

Careful attention has been paid to this article series: https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

In particular we do not pull down the changes from the other branch, we instead review the patch to make sure it's safe. We also guard against timing attacks (I think!) by checking the diff hash matches, and only merging when the PR hash matches.

I believe that the `github.token` is limited to the repository, so the scope of this going wrong is, I think, limited. I (and probably others) have backups of the git history on our machines, so I'd like to think the worst that could happen in the event of a vulnerability is the deletion of issues/PRs/etc (which would be bad, but not devastating) and potentially the leaking of any organization global secrets stored in GitHub, which currently is just the codecov token so should not be a huge concern.

Expect some iteration.

### DIFF
--- a/.github/workflows/wgutils-automerge.yml
+++ b/.github/workflows/wgutils-automerge.yml
@@ -11,6 +11,7 @@ permissions:
 
 jobs:
   validate-and-merge:
+    if: ${{ github.event.pull_request.base.ref == 'main' }}
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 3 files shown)


mjmahone (approved):

benjie:
Here goes... :grimac

PR #1618, opened 2024-12-19
This deletion should not be auto-merged

benjie:


benjie:
Didn't get auto-merged; merging manually...

PR #1649, opened 2025-02-15
Add SemanticNullability RFC document

benjie:


benjie:
RFC documents tend to get merged without review.

PR #1650, opened 2025-02-16
I missed the most important part - disabling error propagation!

benjie:
These are all solutions assuming (in general) that error propagation is disabled\*. In general, a client like Relay doesn't want null propagation _anywhere_, so lets take a shared assumption that clients can opt in to having null propagation disabled.

\* _technically_ someone could build a schema using only semantic non-null and no traditional non-null types, and thus clients wouldn't need to opt-in to disabling error propagation since there wouldn't be any, but that would rule out almost every existing GraphQL schema that's in usage.

### DIFF
--- a/rfcs/SemanticNullability.md
+++ b/rfcs/SemanticNullability.md
@@ -109,11 +109,35 @@ to handle errors in the relevant places.
 
 However, this would mean that clients such as Relay would want to add `?` in
 every position, causing an "explosion" of question marks, because really what
-Relay desired was to disable null propagation entirely.
+Relay desired was to disable error propagation entirely.
 
-## Semantic nullability
+## Disabling error propagation
 
-What we ultimately realised is that GraphQL is missing a type.
+It became clear that disabling error propagation was desired by advanced GraphQL
+clients and vital for ensuring that normalized caches were as useful as possible
+and that we could live up to the promise of GraphQL's partial success without
+compromise. But that was only part of the problem - the other part was that we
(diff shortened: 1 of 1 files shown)


rfcs/SemanticNullability.md
```diff
@@ -109,11 +109,35 @@ to handle errors in the relevant places.
...
+`@behavior(onError: null)`; or via a request-level flag) - in general the
+specific mechanism is unimportant and thus solutions are not expected to comment
+on it unless the choice is significant to the proposal.
+
+### Semantic nullability
+
+We realised that if we were to do this, we would need two schemas: one for when
+null bubbling is disabled, where the true nullability of fields could be
+represented; and one for the traditional error handling behavior, where
+nullability would need to factor in that errors can occur.
```
twof, line 135: Can we say why we would need two schemas?

benjie: That's what I was intending to say here... Did I miss the mark? Feel free to submit a clearer edit!

Essentially:

1. we need the existing schema with the existing best practices for when error propagation is enabled, for the reasons that it was originally designed that way and has been successful for 10 years;
2. if we disable error bubbling then we can place "non-null" in more places, forming a second schema (or a second "view" of the schema, if you prefer)

Might be clearer to think of this as "views" of the schema rather than being two distinct schemas. If you look at the two schema views in isolation they're the ideal situation for each type of consumer (bubbling vs non-bubbling); thus our aim is to enable that two-schema-like feel without actually having to maintain two separate schemas - hence semantic nullability.

PR #1652, opened 2025-02-16
Add `@onError` as a possible solution to semantic nullability

martinbonnin:
`@onError` goes full cycle by focusing on disabling error propagation, which was one of the orginal issues sparkling the nullability discussion. It solves semantic nullability by allowing to focus schema design on error-handling clients. 

See https://github.com/graphql/nullability-wg/discussions/85 for the discussion

### DIFF
--- a/rfcs/SemanticNullability.md
+++ b/rfcs/SemanticNullability.md
@@ -189,9 +189,9 @@ Guiding Principles. The scores are:
 The promise of this RFC - the reflection of the semantic nullability of the
 fields.
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] |
-| --------------- | --------------- | --------------- | --------------- |
-| ✅              | ✅              | ✅              | ✅              |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|
+| ✅               | ✅               | ✅               | ✅               | ✅               |
 
 Criteria score: 🥇
 
@@ -201,9 +201,9 @@ Users should be able to adopt semantic nullability into an existing schema, and
 when doing so all existing operations should remain valid, and should have the
 same meaning as they always did.
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] |
-| --------------- | --------------- | --------------- | --------------- |
-| ✅              | 🚫              | ✅              | ✅              |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|
+| ✅               | 🚫              | ✅               | ✅               | ✅               |
 
 Criteria score: 🥇
 
@@ -213,9 +213,9 @@ GraphQL has been public for 10 years and there's a lot of content out there
 noting that GraphQL types are nullable by default (unadorned type is nullable)
 and that `!` means non-nullable. Our changes should not invalidate this content.
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] |
-| --------------- | --------------- | --------------- | --------------- |
-| ✅              | 🚫              | ✅              | 🚫              |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|
+| ✅               | 🚫              | ✅               | 🚫              | ✅               |
 
 Criteria score: 🥈
 
@@ -225,9 +225,9 @@ The GraphQL languages similarity to JSON is one of its strengths, making it
 immediately feel familiar. Syntax used should feel obvious to developers new to
 GraphQL.
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] |
-| --------------- | --------------- | --------------- | --------------- |
-| 🚫              | ✅              | ✅              | ✅              |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|
+| 🚫              | ✅               | ✅               | ⚠️              | ✅               |
 
 Criteria score: 🥈
 
@@ -237,9 +237,9 @@ When a user wishes to replace the value for an input field or argument with a
 variable in their GraphQL operation, the type syntax should be either identical
 or similar, and should carry the same meaning.
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] |
-| --------------- | --------------- | --------------- | --------------- |
-| ✅              | ✅              | ✅              | 🚫              |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|
+| ✅               | ✅               | ✅               | 🚫              | ✅               |
 
 Criteria score: 🥇
 
@@ -248,9 +248,9 @@ Criteria score: 🥇
 Where a proposal allows alternative syntaxes to be used, the two syntaxes should
 not cause confusion.
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] |
-| --------------- | --------------- | --------------- | --------------- |
-| ✅              | ✅              | ✅              | 🚫              |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|
+| ✅               | ✅               | ✅               | 🚫              | ✅               |
 
 Criteria score: 🥇
 
@@ -262,9 +262,9 @@ Template for new items:
 
 DESCRIPTION
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] |
-| --------------- | --------------- | --------------- | --------------- |
-| ❔              | ❔              | ❔              | ❔              |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
+| --------------- | --------------- | --------------- | --------------- |-----------------|
+| ?               | ?               | ?               | ?               | ?               |
 
 Criteria score: ❔
 
@@ -466,3 +466,46 @@ directive is present, and a `?` symbol is used to indicate a nullable position.
 - [F][criteria-f]
   - 🚫 `Int` being nullable in one mode and non-nullable in the other mode is
     unexpected and will likely lead to confusion.
+
+## 💡 5. `@onError` directive
+
+**Champion**: @martinbonnin
+
+- Discussion: https://github.com/graphql/nullability-wg/discussions/85
+
+Instead of introducing a new type, this proposal introduces a new client directive allowing to control the error behaviour and opt-out of error propagation. 
+
+With `@onError(action: NULL)`, schema authors can optimize for error-handling clients and start using `!` to indicate semantically non-null fields.  
+
+```graphql
+enum __ErrorAction {
+  """
+  Non-nullable positions that error cause the error to propagate to the nearest nullable
+  ancestor position. The error is added to the "errors" list.
+  """
+  PROPAGATE
+
+  """
+  Positions that error are replaced with a `null` and an error is added to the "errors"
+  list.
+  """
+  NULL
+}
+
+directive @onError(action: __ErrorAction) on QUERY | MUTATION | SUBSCRIPTION
+```
+
+### ⚖️ Evaluation
+
+- [A][criteria-a]
+  - ✅
+- [B][criteria-b]
+  - ✅ The default `@onError` action is `PROPAGATE`, making sure existing documents are not impacted.
+- [C][criteria-c]
+  - ✅ `Int` means nullable, and `Int!` means non-nullable, still.
+- [D][criteria-d]
+  - ⚠️ Adding `@onError` to operations is not immediately intuitive but most error-handling clients should add it automatically, making it transparent to end users. 
+- [E][criteria-e]
+  - ✅ Same syntax.
+- [F][criteria-f]
+  - ✅ Same syntax.


rfcs/SemanticNullability.md
```diff
@@ -189,9 +189,9 @@ Guiding Principles. The scores are:
 The promise of this RFC - the reflection of the semantic nullability of the
 fields.
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] |
-| --------------- | --------------- | --------------- | --------------- |
-| ✅              | ✅              | ✅              | ✅              |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|
+| ✅               | ✅               | ✅               | ✅               | ✅               |
```
benjie, line 194:
```suggestion
| ✅               | ✅               | ✅               | ✅               | 🚫                |
```

Solution 5 makes it so that there _are_ no positions that can only be null if an error occurred - it doesn't give us the ability to express a position that will only be null if an error occurred.

rfcs/SemanticNullability.md
```diff
@@ -201,9 +201,9 @@ Users should be able to adopt semantic nullability into an existing schema, and
 when doing so all existing operations should remain valid, and should have the
 same meaning as they always did.
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] |
-| --------------- | --------------- | --------------- | --------------- |
-| ✅              | 🚫              | ✅              | ✅              |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|
+| ✅               | 🚫              | ✅               | ✅               | ✅               |
```
benjie, line 206:
```suggestion
| ✅               | 🚫              | ✅               | ✅               | ⚠️               |
```

Perhaps the wording of this solution criteria needs tweaking (or maybe we need a new solution criteria for it) but existing documents are significantly impacted by this proposal - they will no longer be capable of handling partial success, and that feels to me like their meaning has thus changed significantly.

rfcs/SemanticNullability.md
```diff
@@ -225,9 +225,9 @@ The GraphQL languages similarity to JSON is one of its strengths, making it
 immediately feel familiar. Syntax used should feel obvious to developers new to
 GraphQL.
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] |
-| --------------- | --------------- | --------------- | --------------- |
-| 🚫              | ✅              | ✅              | ✅              |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|
+| 🚫              | ✅               | ✅               | ⚠️              | ✅               |
```
benjie, line 230:
```suggestion
| 🚫              | ✅               | ✅               | ✅              | ⚠️               |
```

You updated the wrong column :)

rfcs/SemanticNullability.md
```diff
@@ -466,3 +466,46 @@ directive is present, and a `?` symbol is used to indicate a nullable position.
 - [F][criteria-f]
   - 🚫 `Int` being nullable in one mode and non-nullable in the other mode is
     unexpected and will likely lead to confusion.
+
+## 💡 5. `@onError` directive
```
benjie, line 470: All of these solutions would be paired with an `@onError` or similar directive; and your proposal doesn't really require a specific directive be used - just that null bubbling can be disabled.

```suggestion
## 💡 5. Use non-null in more places and encourage disabling error propagation
```

rfcs/SemanticNullability.md
```diff
@@ -466,3 +466,46 @@ directive is present, and a `?` symbol is used to indicate a nullable position.
...
+
+  """
+  Positions that error are replaced with a `null` and an error is added to the "errors"
+  list.
+  """
+  NULL
+}
+
+directive @onError(action: __ErrorAction) on QUERY | MUTATION | SUBSCRIPTION
+```
```
benjie, line 496:
```suggestion
This proposal relies on the ability of clients to opt out of error propagation; instead of introducing a new type it instructs schema authors to optimize for error-handling clients and use the traditional non-null type (`!`) on all semantically non-null fields.
```

rfcs/SemanticNullability.md
```diff
@@ -466,3 +466,46 @@ directive is present, and a `?` symbol is used to indicate a nullable position.
...
+
+### ⚖️ Evaluation
+
+- [A][criteria-a]
+  - ✅
+- [B][criteria-b]
+  - ✅ The default `@onError` action is `PROPAGATE`, making sure existing documents are not impacted.
```
benjie, line 503:
```suggestion
- [A][criteria-a]
  - 🚫 It's not possible to indicate a position that can be null only on error.
- [B][criteria-b]
  - ⚠️ Existing documents will not disable null propagation and the change in nullability is backwards compatible; however since the schema uses non-null without regard to error production, existing documents will no longer be able to gracefully handle errors that happen in semantically non-nullable positions and thus their meaning has changed.
```

benjie (changes requested):
The ability to disable error propagation is already an assumption of most solutions (though technically they can be used without it, for the full benefit error propagation should be disabled), and I don't think the _specific method_ in which error propagation is disabled is important to your proposal. Hopefully I've captured your intent in the edits :+1:

rfcs/SemanticNullability.md
martinbonnin: I have to disagree here. Solution #5 is saying that if I use `onError(action: NULL)` and I see this in the schema:

```graphql
type User {
  email: String!
}
```

Then I can assume that `data.user.email` in the response is never null unless there is also a matching error in the errors array. What am I missing here?

rfcs/SemanticNullability.md
martinbonnin: If an existing executable document doesn't use `@onError(action: NULL)` then nothing changes. It will receive the same response as it used to?

rfcs/SemanticNullability.md
martinbonnin: Woopsie! Thanks

rfcs/SemanticNullability.md
martinbonnin: Fair enough. Feels like it would help the discussion to give each proposal a memorable, short name (and not only numbers) but we can discuss that later.

rfcs/SemanticNullability.md
martinbonnin: Alright, I just read https://github.com/graphql/graphql-wg/pull/1652#discussion_r1957294407 and I understand better now but still think I disagree. Whether there are changes for existing document is not dependant on the proposal, it's on API providers. 

Some APIs providers may decide to keep optimizing for non-error-aware client as long as they want to. The proposal in itself doesn't change the meaning of existing documents. Only when (if) a provider switches to a non-nullable schema does that happen.

rfcs/SemanticNullability.md
benjie: That relies on the client disabling null bubbling, in which case all non-nulls become semantically non-null as they do in every solution. For clients not disabling null bubbling there is no position which is indicated to be null only on error, since every position is either nullable, or strictly non-nullable. I.e. this solution does not implement the “two schemas” described in the intro, it does not introduce a semantic non-null type. Put another way, the fields you say are “null only on error” with null bubbling turned off are NOT null “null only on error” with null bubbling on, they are “never null”. “Null only on error” / semantic non-null is a property that should hold in both modes.

rfcs/SemanticNullability.md
benjie: If a field isn’t expected to ever throw errors and isn’t semantically nullable then there’s no particular reason it shouldn’t already be non-nullable in existing schemas, so the proposal only impacts semantically-non-nullable positions that are currently nullable because they are expected to potentially throw errors. Either your solution is telling people to use non-null in these semantically non-null positions or it isn’t. If it is, then it’s a change in schema types that will lead to existing deployed clients managing errors less gracefully. If it isn’t then it’s essentially status quo with the ability to disable null bubbling, which is a separate and important concern, but it doesn’t solve the “semantic nullability” problem.

rfcs/SemanticNullability.md
martinbonnin: > this solution does not implement the “two schemas” described in the intro

"two schema" is a possible solution to the [larger problem](https://github.com/graphql/graphql-wg/blob/main/rfcs/SemanticNullability.md#-problem-statement) of `GraphQL needs to be able to represent semantically nullable and semantically non-nullable types as such when error propagation is disabled.` 

Criteria A. is `GraphQL should be able to indicate which positions will only be null if an error occurred` and proposal `#5` qualifies for error-handling clients and GraphQL schemas that decide to switch to non-null by default. This population is a subset of the general GraphQL population but it's still a way forward.

benjie: Right, but by only applying to a portion of the audience it does not meet the criteria - there is no type that represents "null iff error". That's okay - your solution doesn't need to meet all of the criteria, it's just a comparison! None of the solutions meet all of the criteria.

If you disable null bubbling then _all_ non-nulls pass this criteria; if you want to add another criteria that says this then please feel free, but this one should apply to all users not just those with null bubbling disabled. Alternatively you should propose a solution where the schema disables null bubbling for everyone (not based on client opt-in) - that would pass this criteria.

rfcs/SemanticNullability.md
benjie: > Perhaps the wording of this solution criteria needs tweaking (or maybe we need a new solution criteria for it) but existing documents are significantly impacted by this proposal

I've moved this concern to a separate criteria so it can be analysed separately:

https://github.com/graphql/graphql-wg/pull/1654

You can close this conversation if you like :)

rfcs/SemanticNullability.md
martinbonnin: (Thanks for looking into this! Answering here to keep the discussion in one place.)

For some context, the new criteria now reads:

```
GraphQL should be able to indicate which nullable fields should become non-nullable 
when error propagation is disabled
``` 

I'm okay with this being a criteria but don't think it should be the first, must-have criteria because it disqualifies other approaches like `#5` (and maybe others).

May I suggest something more general like:

```
Criteria A. Developers should be able to tell, by looking at the schema, 
whether a position can have a semantic null value.
```

It doesn't have to be all apply to all developers. If it allows 90% of them to remove null checks, it's already something. 


I've just seen https://github.com/graphql/graphql-wg/pull/1654. It's okay for `#5` to not pass Criteria G. but I think it should pass #1.

martinbonnin: For the record, I would also challenge Criteria G. as a "must-have". Software evolves and it might be acceptable to keep some legacy APIs behind. To quote the Kotlin evolution principles: 

```
Language design is cast in stone,
but this stone is reasonably soft,
and with some effort we can reshape it later.
```

benjie: Perhaps the title of the proposal itself is wrong then! Unfortunately `WhichNullableFieldsShouldBecomeNonNullableWhenErrorPropagationIsDisabled.md` felt a bit long for the filename :wink:

martinbonnin:
@benjie I have merged your edits and added an objection about Criteria G. Can you merge this PR if it works with you?

rfcs/SemanticNullability.md
```diff
@@ -523,3 +526,28 @@ directive is present, and a `?` symbol is used to indicate a nullable position.
...
+- Discussion: https://github.com/graphql/nullability-wg/discussions/85
+
+This proposal relies on the ability of clients to opt out of error propagation; instead of introducing a new type it instructs schema authors to optimize for error-handling clients and use the traditional non-null type (`!`) on all semantically non-null fields.
+
+### ⚖️ Evaluation
+
+- [A][criteria-a]
+  - 🚫 It's not possible to indicate a position that can be null only on error.
+- [B][criteria-b]
+  - ⚠️ Existing documents will not disable null propagation and the change in nullability is backwards compatible; however since the schema uses non-null without regard to error production, existing documents will no longer be able to gracefully handle errors that happen in semantically non-nullable positions and thus their meaning has changed.
```
benjie, line 543:
```suggestion
  - ✅ The change from nullable to non-nullable on output is backwards compatible from a type perspective; for impact on error boundaries see G.
```

rfcs/SemanticNullability.md
martinbonnin: Maybe `SemanticNullabilityTypeSystem.md` instead of `SemanticNullability.md`? Changing the type system is one way amongst others to achieve semantic nullability?

rfcs/SemanticNullability.md
```diff
@@ -523,3 +526,28 @@ directive is present, and a `?` symbol is used to indicate a nullable position.
...
+**Champion**: @martinbonnin
+
+- Discussion: https://github.com/graphql/nullability-wg/discussions/85
+
+This proposal relies on the ability of clients to opt out of error propagation; instead of introducing a new type it instructs schema authors to optimize for error-handling clients and use the traditional non-null type (`!`) on all semantically non-null fields.
+
+### ⚖️ Evaluation
+
+- [A][criteria-a]
+  - 🚫 It's not possible to indicate a position that can be null only on error.
```
benjie, line 541:
```suggestion
  - 🚫👍 The nullability used in both error-propagation and no-error-propagation modes are the same. This is a feature, not a bug!
```

rfcs/SemanticNullability.md
```diff
@@ -224,9 +224,9 @@ Note: Traditional non-nullable types will effectively become semantically
 non-nullable when error propagation is disabled no matter which solution is
 chosen, so this criteria is only concerned with traditionally nullable types.
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] |
-| --------------- | --------------- | --------------- | --------------- |
-| ✅              | ✅              | ✅              | ✅              |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|
+| ✅               | ✅               | ✅               | ✅               | 🚫                |
```
benjie, line 229:
```suggestion
| ✅               | ✅               | ✅               | ✅               | 🚫👍              |
```

benjie (changes requested):
@martinbonnin I've used :no_entry_sign: :+1:  to indicate that though it doesn't pass the criteria, this as seen as a positive; I also updated criteria B to a pass.

benjie:
Please feel free to follow up with more PRs! I was keen to get this merged :)

martinbonnin:
@benjie sounds good, thanks!

rfcs/SemanticNullability.md
benjie: I think "add semantic nullability" is sufficiently descriptive - with your proposal we don't _need_ semantic nullability to be added - GraphQL already possesses what it needs. And "add" is kind of implicit in most RFCs: "add input unions", "add fragment arguments", "add schema metadata", "add a struct type", etc.

martinbonnin: > with your proposal we don't need semantic nullability to be added - GraphQL already possesses what it needs.

It possesses what it needs from a type system point of view but the "nullable by default" best practice voids that because the semantic information is lost during implementation (or else you lose partial data). 

I would argue that GraphQL will support semantic nullability when it's both in the type system and in the best practices. Contrary to the other proposals, `#5` is not touching the type system but is touching the best practices to achieve end to end semantic nullability (and that's possible because of no-error-propagation).

Anyways, not a huge deal, we can keep the filename as is for now and leave it some time to settle.

PR #1656, opened 2025-02-17
Add implementation and spec simplicity

twof:
This has come up as a decision point in discord a few times. In my opinion, the proposal (and GQL generally) should operate in service of the end user, and we should do what's necessary to fulfill user needs. If the spec or the implementation cannot fulfill user needs, then changes should be made so that they can.


benjie (approved):
I'd argue that if you ignore document directives which are an external concern, (3) is also pretty simple both to specify and to implement. That said, the fact it requires document directives in itself makes it complex, so I'm inclined to agree :+1:

PR #1657, opened 2025-02-17
! meaning compatability updated

twof:
In option 3, the meaning and mechanics of a `!` adorned type change when a document adopts semantic nullability

### DIFF
--- a/rfcs/SemanticNullability.md
+++ b/rfcs/SemanticNullability.md
@@ -250,7 +250,7 @@ and that `!` means non-nullable. Our changes should not invalidate this content.
 
 | [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] |
 |-----------------|-----------------|-----------------|-----------------|-----------------|
-| ✅               | 🚫              | ✅               | 🚫              | ✅               |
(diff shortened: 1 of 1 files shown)


benjie:
Criteria C has been clarified to only relate to the unadorned type; if you wish to add that `!` should only mean _strictly_ non-nullable in all modes then you should do so with a new criteria.

rfcs/SemanticNullability.md
```diff
@@ -497,10 +497,10 @@ day-to-day work.
     input and output (the difference between semantic an strict non-null does
     not occur on input)
 - [F][criteria-f]
-  - ✅ `Int` reatains its meaning across both modes, and `Int!` means
-    non-nullable in both modes. Only the SDL ever uses `Int!!` and it still
-    means non-null, just with the additional "kills parent on exception"
-    behavior.
+  - 🚫 `Int` reatains its meaning across both modes, but `Int!` means
+    non-nullable in one mode, but semantic non-nullable in another. Only
```
benjie, line 501: "modes" here was referring to null bubbling enabled/disabled, and `Int!` always means semantic non-null when null bubbling is disabled, so all solutions would fail this criteria the same. As such, the commentary around `Int!` meaning has been removed from the criteria.

PR #1658, opened 2025-02-17
Remove criteria F

twof:
Criteria F is a duplicate of Criteria C and is overly vague

### DIFF
--- a/rfcs/SemanticNullability.md
+++ b/rfcs/SemanticNullability.md
@@ -278,17 +278,6 @@ or similar, and should carry the same meaning.
 
 Criteria score: 🥇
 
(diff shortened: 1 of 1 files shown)


benjie:
F is not a duplicate of C, which is demonstrated by the evaluations for them differing - both proposal 2 and 4 fail criteria C (clarified in #1659), but for F proposal 2 is consistent because in a new schema with `@strictNullability`, `Int` always means "semantically non-nullable". Proposal 4 fails criteria F because it has two modes of operation on the same schema, and `Int` means essentially opposite things in these two modes: either nullable or non-nullable depending on the mode. Perhaps rather than removing this you can help reword it to make the difference clearer?

martinbonnin (approved):

PR #1661, opened 2025-02-18
SemanticNullability -> SemanticNullabilityType; add precision

benjie:
Following discussion with @martinbonnin it became obvious that the title of this RFC was too broad and seemed to make it such that any proposal to disable null bubbling would solve the problem; however as outlined in the problem history the intent of this RFC is that schemas gain a new type to enable splitting the existing "nullable" into "semantically nullable" and "semantically non-nullable". I've renamed the RFC to include the word "type" and have added precision to the problem statement so that it better outlines the problem it is attempting to solve and is a better summary of the problem history section above it.

With this change, I think we should explicitly indicate that solution 5 is a counter-proposal - it's an argument against the need for a new explicit semantic nullability type.

cc @martinbonnin @twof

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,3 @@
 agendas/*/*.md
 agendas/*/*/*.md
-rfcs/SemanticNullability.md
+rfcs/SemanticNullabilityType.md
--- a/rfcs/SemanticNullabilityType.md
+++ b/rfcs/SemanticNullabilityType.md
@@ -1,4 +1,4 @@
-# RFC: Semantic Nullability
+# RFC: Semantic Nullability Type
 
 # 📜 Problem History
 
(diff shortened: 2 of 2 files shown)


benjie:
Closing as I'm inclined to agree with Martin that "semantic nullability" is the problem we're trying to solve for, and solution 5 - namely: use strictly-non-null to represent semantic-non-null - is a solution to that.

martinbonnin:
[Link to the Discord discussion for future reference](https://discord.com/channels/625400653321076807/971515695881924621/1341344425179091028):

``` 
I think the original focus on "semantic nullability" was the good one actually. This is the problem we want to solve. Whether we do it with new types or GraphQL 2.0 depends where we want to put the tradeoff but focusing on type system feels like narrowing the scope of possible solutions
```

PR #1667, opened 2025-02-21
Add objection: lower the score of `The solution should add value even with error propagation enabled` 

martinbonnin:
Feels like this is outside the "semantic nullability" problem. 

The problem of indicating whether or not a position can error in the schema is a separate, larger problem IMO that could require larger changes like adding even more types to the type system:

```rust 
String
Option<String>
Result<String, Error>
Result<Option<String>, Error> 
```

This is a much wider question though and I suggest we leave it out for now.


benjie:
The score was lowered in #1686

PR #1670, opened 2025-02-22
Simplify and standardize syntax comparison

benjie:
Let's just look at the input/output syntax differences rather than having massive tables. If a syntax doesn't have a "&rArr;" then it's unchanged versus the current (traditional) syntax.


benjie:
It's not clear to me what the input syntax is that Lee is recommending. I've taken a guess that @twof is proposing `Int?` for nullable on input and `Int` for non-nullable on input (rather than requiring `Int?` and `Int!` and not allowing unadorned on input?) - please send a PR if I got this wrong!

PR #1706, opened 2025-03-28
Fold rejected solutions for easier processing

martinbonnin:
For simplicity, I marked the solutions as "rejected" as we need to move forward. If you believe your solution was unfairly rejected, please make an appeal here!


benjie:
Can we leave the title outside of the `<details>` and just put `<summary>Rejected</summary>` inside? The advantage of this is that we can still link to the individual solutions for people to easily expand them and read their history.

benjie (approved):

martinbonnin:
Merging so I can rebase https://github.com/graphql/graphql-wg/pull/1707

PR #1707, opened 2025-03-28
Add solution 7

martinbonnin:
This adds [@leebyron's solution](https://github.com/graphql/graphql-wg/discussions/1700) to the comparison


rfcs/SemanticNullability.md
```diff
@@ -1005,3 +1005,63 @@ positions.
...
+
+## 💡 7. `@propagateError` directive
+
+[solution-7]: #-7-propagateerror-directive
+
+**Champion**: @leebyron
+
+Discussion: https://github.com/graphql/graphql-wg/discussions/1700
+
+This proposal introduces a directive that can be added to fields to indicate that they should propagate errors.
```
benjie, line 1017:
```suggestion
This proposal changes the `!` symbol and the `NON_NULL` introspection kind both to mean "semantic non null" (allowing for `null` on error). It introduces the `@propagateError` directive that can be added to fields to indicate that they should propagate errors in order to provide backwards compatibility with existing deployed clients.
```

rfcs/SemanticNullability.md
```diff
@@ -464,9 +464,9 @@ Per Benoit:
 > an outcome of this whole effort I’d like to see happening.
 
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] | [6][solution-6] |
-|-----------------|-----------------|-----------------|-----------------|-----------------|-----------------|
-| ✅               | ✅              | ✅               | ✅               | 🚫               | ⚠️           |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] | [6][solution-6] | [7][solution-7] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|-----------------|-------------|
+| ✅               | ✅              | ✅               | ✅               | 🚫               | ⚠️           | ✅            |
```
benjie, line 469: If `propagateErrorOnAllNonNullFields: true` is set, then adding `!` to more fields changes the error boundaries, and thus means that existing users cannot add `!` in more places without breaking existing clients error resilience.

Setting `propagateErrorOnAllNonNullFields: false`, even once all the fields are marked up with `@propagateError`, is a breaking change from introspection point of view so cannot be done until all clients and tooling have been updated to understand the new `NON_NULL` meaning and the additional introspection field. As such, `propagateErrorOnAllNonNullFields: false` cannot be safely adopted in a public API, partner API, or any other API where you do not control all clients and tooling.

Thus I'd argue that this solution fails P except for private APIs (and even there, the migration story is painful if you don't have teams dedicated to it).

rfcs/SemanticNullability.md
```diff
@@ -311,9 +311,9 @@ still keep errors local to the same positions that they did when they were
 published), allowing for the "partial success" feature of GraphQL to continue to
 shine and not compromising the resiliency of legacy deployed app versions.
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] | [6][solution-6] |
-| --------------- | --------------- | --------------- | --------------- |-----------------|-----------------|
-| ✅              | ✅              | ✅              | ✅              | 🚫              | ✅          |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] | [6][solution-6] | [7][solution-7] |
+| --------------- | --------------- | --------------- | --------------- |-----------------|-----------------|-------------|
+| ✅              | ✅              | ✅              | ✅              | 🚫              | ✅          | ✅            |
```
benjie, line 316:
```suggestion
| ✅              | ✅              | ✅              | ✅              | 🚫              | ✅          | ⚠️            |
```

With `propagateErrorOnAllNonNullFields: true`, adding `!` (semantic non-null) in existing positions will change error propagation boundaries in existing (deployed) executable documents, in the same way that it would for solution 6, so semantic non-null could not be added until `propagateErrorOnAllNonNullFields` is set to `false` (i.e. after all clients and tooling migrate).

rfcs/SemanticNullability.md
```diff
@@ -367,9 +367,9 @@ Criteria score: 🥇
 
 We do not want to break existing tooling.
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] | [6][solution-6] |
-|-----------------|-----------------|-----------------|-----------------|-----------------|-----------------|
-| ✅               | ❔              | ✅               | ❔               | ✅               | ✅          |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] | [6][solution-6] | [7][solution-7] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|-----------------|-------------|
+| ✅               | ❔              | ✅               | ❔               | ✅               | ✅          | ✅            |
```
benjie, line 372:
```suggestion
| ✅               | ❔              | ✅               | ❔               | ✅               | ✅          | ⚠️            |
```

This solution is non-breaking _if_ you keep `propagateErrorOnAllNonNullFields: true` enabled forever.

Otherwise, it is a breaking change as it changes the meaning of `NON_NULL` in the schema, preventing adoption until all clients and tooling have been updated.

rfcs/SemanticNullability.md
```diff
@@ -435,9 +435,9 @@ Per Lee:
 > allow inconsequential changes in behavior, but bubbling the error up isn't
 > inconsequential.)
 
-| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] | [6][solution-6] |
-|-----------------|-----------------|-----------------|-----------------|-----------------|-----------------|
-| ✅               | ❔              | ✅               | ✅               | 🚫               | ✅          |
+| [1][solution-1] | [2][solution-2] | [3][solution-3] | [4][solution-4] | [5][solution-5] | [6][solution-6] | [7][solution-7] |
+|-----------------|-----------------|-----------------|-----------------|-----------------|-----------------|-------------|
+| ✅               | ❔              | ✅               | ✅               | 🚫               | ✅          | ✅            |
```
benjie, line 440:
```suggestion
| ✅               | ❔              | ✅               | ✅               | 🚫               | ✅          | ⚠️            |
```

Unless `propagateErrorOnAllNonNullFields` is set to `true`, any existing executable documents that introspect types will receive a change in meaning for `NON_NULL`.

benjie (commented):
I've commented on a few of the criteria and have added more detail to the description

rfcs/SemanticNullability.md
martinbonnin: Updated in https://github.com/graphql/graphql-wg/pull/1707/commits/a85fbab15a56ca5684f3f6bfd95d1f4bf54c1a5c

martinbonnin:
Should we merge this one? I'm fine if we tweak the criteria, etc.. but feels like 7. should be present in the document at least?

PR #1708, opened 2025-03-28
Update solution 6.

martinbonnin:
Update solution 6. As I understand, it also relies on a request parameter as well as introspection changes.

Also added my name as a champion here since it was empty. If anyone feels strongly about this solution, I'm happy to relinquish the "championability"!

### DIFF
--- a/rfcs/SemanticNullability.md
+++ b/rfcs/SemanticNullability.md
@@ -940,13 +940,17 @@ This proposal relies on the ability of clients to opt out of error propagation; 
 
 [solution-6]: #-6-semanticnonnull-directive
 
-**Champion**: -
+**Champion**: @martinbonnin
 
 Outline: https://specs.apollo.dev/nullability/v0.4/#@semanticNonNull
 
-This proposal (which is already adopted in a few places!) introduces a
-directive that can be added to fields to indicate their semantic nullability
-(and that of their nested list positions).
+This proposal relies on:
+- A request parameter to disable error propagation
+- A new directive, `@semanticNonNull`
+- A new introspection argument, `includeSemanticNonNull` (`__Field.type(includeSemanticNonNull: Boolean! = false)`)
+
+The directive (which is already adopted in a few places!) can be added to 
+fields to indicate their semantic nullability(and that of their nested list positions).
 
 ```graphql
 directive @semanticNonNull(levels: [Int!]! = [0]) on FIELD_DEFINITION


rfcs/SemanticNullability.md
```diff
@@ -940,13 +940,17 @@ This proposal relies on the ability of clients to opt out of error propagation;
...
-**Champion**: -
+**Champion**: @martinbonnin
 
 Outline: https://specs.apollo.dev/nullability/v0.4/#@semanticNonNull
 
-This proposal (which is already adopted in a few places!) introduces a
-directive that can be added to fields to indicate their semantic nullability
-(and that of their nested list positions).
+This proposal relies on:
+- A request parameter to disable error propagation
```
benjie, line 948: Disabling error propagation is an orthogonal concern - the solutions benefit the most when both are implemented, but neither relies on the other.

```suggestion
```

rfcs/SemanticNullability.md
```diff
@@ -940,13 +940,17 @@ This proposal relies on the ability of clients to opt out of error propagation;
...
+- A new directive, `@semanticNonNull`
+- A new introspection argument, `includeSemanticNonNull` (`__Field.type(includeSemanticNonNull: Boolean! = false)`)
```
benjie, line 950:
```suggestion
- Either:
  - A new introspection argument, `includeSemanticNonNull` (`__Field.type(includeSemanticNonNull: Boolean! = false)`), if represented as a type, or
  - A new introspection field, `__Field.semanticNonNullLevels: [Int!]`, if represented as field metadata
```

rfcs/SemanticNullability.md
```diff
@@ -940,13 +940,17 @@ This proposal relies on the ability of clients to opt out of error propagation;
...
+
+The directive (which is already adopted in a few places!) can be added to 
+fields to indicate their semantic nullability(and that of their nested list positions).
```
benjie, line 953:
```suggestion
fields to indicate their semantic nullability (and that of their nested list positions).
```

benjie (changes requested):

benjie:
I've been acting champion of this (e.g. [here](https://github.com/graphql/graphql-wg/discussions/1700#discussioncomment-12634500)), and if it's the solution that we select then I plan to implement it on top of the work I've already done. To me, `*` and `@semanticNonNull` are variants of the same proposal, only really differing in terms of the syntax through which they're exposed. That said, if we get it down to `*` vs `@semanticNonNull` I will be arguing for `*` so it makes sense for you to champion `@semanticNonNull` against me if you truly feel it's the superior solution :+1:

benjie (approved):

PR #1729, opened 2025-04-25
Add field extensions RFC

martinbonnin:


benjie:
RFC documents get merged without much in the way of review :+1:

PR #1731, opened 2025-04-25
Adding arguments and enumValues into the mix

martinbonnin:
Also add a "questions" section


benjie:
I would highly recommend that you address these through separate RFCs - just like adding `@deprecated` to different places, each of these positions has their own concerns. I think it's going to be a _lot_ easier to extend a field with directives than it will be to extend an argument with directives, for example.

martinbonnin:
Works for me 👍

PR #1755, opened 2025-05-28
Add `onError` RFC

martinbonnin:
This is a RFC that aims to extract the `onError` bits from the semantic nullability discussion, breaking down the initial problem in smaller, hopefully more manageable, problems.

[Read the rendered RTC](https://github.com/graphql/graphql-wg/blob/0e821e8aa9f881252b4dcb5f4218f7de1a17367a/rfcs/OnError.md)

### DIFF
--- a/rfcs/OnError.md
+++ b/rfcs/OnError.md
@@ -41,20 +41,12 @@ In the end state where error propagation is disabled and clients are error-handl
 The default error propagation is indicated in introspection with a new meta-field, `onErrorDefault` of `String!` type.
 
 ```graphql
-type __Service {
+extend type __Schema {
     """
     Indicates the error behaviour when `onError` is not present in the request.
     One of `NULL`, `PROPAGATE`, `HALT`
     """
     onErrorDefault: String!
-}
-extend type __Schema {
-    """
-    Information relative to the service.
-    In a perfect world, this is a separate meta-field separate from __Schema but new meta fields are not discoverable
-    and this is added as a `__Schema` field for backward compatibility reasons.
-    """
-    service: __Service
 }
 ```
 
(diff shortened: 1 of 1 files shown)


rfcs/OnError.md
```diff
@@ -0,0 +1,79 @@
...
+
+Possible values for the `onError` property are:
+
+* `NULL`: disables error propagation. The executor emits `null` in the `data` part of the response and a GraphQL [execution error](https://spec.graphql.org/draft/#execution-error) in the `errors` part of the response.
+* `PROPAGATE`: propagates the error to the nearest nullable parent field. This is the current behaviour.
+* `HALT`: stops executing the request whenever an error is encountered. This allows services to save resources if clients do not consume partial data.
+
+Servers not supporting `onError` must execute in the `PROPAGATE` mode. 
+
+Servers supporting `onError` must decide on a sensible default. Existing servers will most likely use `PROPAGATE`, new servers will most likely use `NULL`. `HALT` is also possible as a default albeit it's expected to be used less frequently.
```
BoD, line 37: Would it be simpler if the default is always `PROPAGATE`? That automatically makes this compatible with old clients, and removes the need to surface the default in introspection (unless there's another reason to do so that I missed?)

rfcs/OnError.md
```diff
@@ -0,0 +1,79 @@
...
+{
+  "query": "{ helloWorld }",
+  "onError": "NULL"
+}
+```
+
+Possible values for the `onError` property are:
+
+* `NULL`: disables error propagation. The executor emits `null` in the `data` part of the response and a GraphQL [execution error](https://spec.graphql.org/draft/#execution-error) in the `errors` part of the response.
+* `PROPAGATE`: propagates the error to the nearest nullable parent field. This is the current behaviour.
```
calvincestari, line 32:
```suggestion
* `PROPAGATE`: propagates `null` and the error to the nearest nullable parent field. This is the current behaviour.
```
I think having only "the error" could be misinterpreted and it's clearer to state that the `null` value is propagated too.

PR #1773, opened 2025-06-19
Reschedule June secondary meeting back a week

benjie:
Due to Juneteenth it looks like attendance will be poor, moving to next week (agreed with Lee).

cc @magicmark


martinbonnin:
This clashes with OTEL and HTTP working groups I think? I'm fine merging those together FWIW, I couldn't do today but next week works 👍 . Just an FYI.

benjie:
Cancelled the HTTP WG, not too worried about OTEL clash, secondary meetings don't tend to be as populated.

PR #1813, opened 2025-09-04
Cancel Sept primary - see you at GraphQLConf!

benjie:


martinbonnin:
I was planning to join only to celebrate the new spec cut 🍾! But we can do that IRL in a few days, see you there everyone.

PR #1817, opened 2025-09-09
RFC: `@matches` directive

magicmark:
This RFC proposes `@matches` as a way for clients to express what types they support when receiving an array of unions of polymorphic types:

**SDL**

```graphql
union Media = Book | Movie | Opera

type Query {
  getMedia(supports: [String!] @matches): [Media]
}
```

**Query**

```graphql
query {
  # This query was written before the Opera type was added.
  # @matches declares that the resolver must only return "supported" types, as specified by the argument.
  getMedia(supports: ["Book", "Movie"]) {
    ... on Book { author }
    ... on Movie { director }
  }
}
```

See the full RFC here: https://github.com/magicmark/graphql-wg/blob/patch-18/rfcs/MatchesDirective.md

(This came out of a GraphQLConf unconf discussion topic - thanks to @captbaritone @rbalicki2 @benjie! :) 

**Open questions:**
- [ ] `@match` vs `@matches` (singular is consistent with `@include`)
- [ ] string parsing is yucky. is this the right syntax?
- [ ] should we make it repeatable? or keep a tighter scope?

### DIFF
--- a/rfcs/MatchesDirective.md
+++ b/rfcs/MatchesDirective.md
@@ -6,7 +6,11 @@
 
 This RFC proposes adding a new directive `@matches` and associated validation
 rules to enforce the safe selection of "supported" types when using fragment
-spreads on a field that returns an array of unions of polymorphic types.
+spreads on a field that returns an array of unions of polymorphic types:
+
+```graphql
+directive @matches(sort: Boolean = True) on ARGUMENT_DEFINITION
+```
 
 ## 📜 Problem Statement
 
--- a/rfcs/MatchesDirective.md
+++ b/rfcs/MatchesDirective.md
@@ -203,7 +203,7 @@ fragment MoviesTab on Query {
 The selection sets in fragments get merged, and the `getMedia` resolver is only
 executed once. With the above implementation, we'd *only* return Books (the
 first to be evaluated). Since the resolver is only executed once, this is
-impossible to implement correclty.
+impossible to implement correctly.
 
 #### 3.b. Add the `supports` argument at runtime
 
--- a/rfcs/MatchesDirective.md
+++ b/rfcs/MatchesDirective.md
@@ -9,7 +9,7 @@ rules to enforce the safe selection of "supported" types when using fragment
 spreads on a field that returns an array of unions of polymorphic types:
 
 ```graphql
-directive @matches(sort: Boolean = True) on ARGUMENT_DEFINITION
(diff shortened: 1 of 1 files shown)


rbalicki2:
FYI the link to the RFC doesn't work

magicmark:
thanks @rbalicki2! thanks for the feedback, will incorporate this.

> I think that it's probably worth considering treating the server validation piece and the operation-validation piece as independent, at least in terms of inclusion in the spec. I of course like both!

Happy to do this in whichever way makes the most sense to most people, but my thinking was this makes the most sense a combined proposal rather than splitting them into independent proposals that might land at different times. I'd be slightly nervous about folks using the new directive without validation rules, since it seems too easy to construct an invalid query by passing in garbage array arguments (with relay this is impossible)

magicmark:
@benjie good point!

How about something like this...

```graphql
type Query {
  allAnimals(
    first: Int
    after: String
    only: [AnimalType!] @matches(path: "nodes")
  ): AnimalConnection
}

type AnimalConnection {
  nodes: [Animal!]
  pageInfo: PageInfo
}
```

And I suppose it's worth making `repeatable` in order to support something (convoluted) like this:

```graphql
type Query {
  petsForSale(
    only: [AnimalType!] @matches(path: "largePets") @matches(path: "bigPets")
  ): StoreFront
}

type StoreFront {
  """
  e.g. Great Danes, Horses etc
  """
  largePets: [Animal!]

  """
  e.g. Gerbils, Siberian Hamsters etc
  """
  bigPets: [Animal!]
}
```

(path could also be an `array`, but I'd vote for to optimize the simpler interface for the more common pattern)

BoD:
@rbalicki2 Interesting! Yeah I think if the plan is for this directive to be part of the main spec, the expected behavior of the server probably needs to be expressed precisely. 

On a different topic, was this considered: a mechanism where the known types are automatically inferred from the list of type conditions:

```graphql
query GetMedia {
  getMedia @onlyMatchingTypesOrAnotherNameToBeDertermined {
    ... on Book { title author }
    ... on Movie { title director }
  }
}
```
?

This would return only Book or Movie. That would avoid having to add a dedicated argument to the field.

benjie:
@BoD How would that work with connections?

```graphql
query GetMedia {
  getMediaConnection @onlyMatchingTypesOrAnotherNameToBeDetermined(path: ["edges", "node"]) {
    edges {
      cursor
      node { 
        ... on Book { title author }
        ... on Movie { title director }
      }
    }
  }
}
```

PR #1841, opened 2025-10-15
RFC: Object Identification

phryneas:
This might be a déjà vu for some, but I want to give this another try.

Note that this is different from the `@key` and `@lookup` directives in the Composite Schemas Spec as I also want to solve this problem for non-schema-aware clients, so while those directives are amazing for federation, they cannot help us solve the use case I'm laying out here.

Those two proposals are more complementary than conflicting in my eyes.


benjie (commented):
RFC documents get merged without review.

martinbonnin:
Might be worth having a matching tracking issue at https://github.com/graphql/graphql-spec/issues/?

PR #1874, opened 2025-12-13
Abstract filter argument

benjie:
@magicmark: please fully replace this description with your own.

---

Read the `README` as the first thing, the spec is just super rough key algorithms and rules.

This PR is a quick (hah!) write up of my thoughts following a chat with @magicmark this morning. I've opened it as a draft with the intent of @magicmark editing it as he sees fit (or replacing it!) and the completing a first draft. It is far from complete, with major gaps, but I hope it helps to capture some of the points discussed and the reasoning behind some of the decisions we're leaning towards.

Please note that the draft spec is NOT fully in spec-like language; such work (especially around talking about connections, and the specifics of returning `[AbstractType]` vs `[AbstractType!]!`) would take a lot longer so I've mostly skipped over it.

@magicmark You may merge this as-is and the follow up with your own edits, or edit it directly in this PR, I don't mind. We're pretty flexible on what gets merged into the RFCs folder. Also the names of things are all placeholders, feel free to replace! Ideally keep the `git` history intact though :)

### DIFF
--- a/rfcs/AbstractFilter/README.md
+++ b/rfcs/AbstractFilter/README.md
@@ -185,7 +185,7 @@ into a connection?
 It's essential that the client know that this filtering is applied such that its
 normalized store does not become corrupted, and thus it was agreed that we
 should stick with passing the list of allowed types as an argument (since
-arguments already factor into npormalized cache keys).
+arguments already factor into normalized cache keys).
 
 ### Two specifications
 


magicmark:
some meta-context - this proposal is not to change the core language of GraphQL

This is indented to be supplementary: a "userland" or "contrib" spec. (Prior art includes the connections and global objects specification).

The idea of submitting this as an RFC is similar to that of Python's [PEPs](https://peps.python.org/pep-0000/#numerical-index), which may cover non-core-language things such as [formatting](https://peps.python.org/pep-0008/).

benjie:
Thanks for writing that up @magicmark; I've written a bit of a proposal here: https://github.com/graphql/graphql-wg/issues/1879 feel free to weigh in on it (even though it's tagged "TSC discussion" it's not private, we have a separate repo for that) with your own thoughts e.g. your PEP-like suggestions.

magicmark:
@benjie i'm working on this now. Any reason this spec should be limited to lists specifically?

> ...or to a field whose return type is a **list** and the named return type is an abstract type.

For example, this seems like a valid use case?

```graphql
interface BannerAd { ... }
type StaticImageAd implements BannerAd { ... }
type InteractiveHTMLCanvasAd implements BannerAd { ... }
type FlashGameAd implements InteractiveBannerAd { ... }

type Query {
  adBanner(only: String @limitTypes): BannerAd
}
```

magicmark:
@benjie thoughts on if this spec should also support a field that returns a single type?

e.g. this could still be useful even when returning a single type

```graphql
interface BannerAd { ... }
type FlashGameAd implements BannerAd { ... }
type InteractiveHTMLCanvasAd implements BannerAd { ... }
type StaticImageAd implements BannerAd { ... }

type Query {
  bannerAd(only: String): BannerAd
}
```

PR #1886, opened 2026-01-12
Add self to attendees

janettec:


martinbonnin:
@janettec Nice!! Glad that you'll be there! :) 

Should I edit the agenda to put your name on the "Fragments are not for reuse" item? I added it a long time ago to make sure it doesn't fall through but you'll have more context.

PR #1989, opened 2026-06-02
Note that emeriti affiliation reflects term

benjie:
Emeriti may have had multiple affiliations during their term.

**Proposal**: one of these should be shown in the affiliation column; once a TSC members term ends we will not keep their affiliation up to date with their latest employment.


martinbonnin (approved):
Thanks for the claryfing note! 

I'd also be happy to remove the column altogether for Emeriti (but that's a bigger change). 

LGTM!

magicmark (approved):

robrichard (approved):

PascalSenn (approved):

michaelstaib (approved):

PR #2005, opened 2026-06-18
Add directive definition support to opt-in features RFC

glen-84:
## Summary

Updates the opt-in features RFC to support directives, now that directives on directive definitions are part of the draft spec.

- Add `DIRECTIVE_DEFINITION` to the `@requiresOptIn` locations (and the docstring), with an example marking a custom directive opt-in.
- List directive definitions alongside the other kinds (field, input field, argument, enum value) in the introspection prose, rather than adding a separate `__Schema`/`__Directive` block — matching how the section already handles arguments and enum values.
- Align the `includeDeprecated` introspection argument with the current draft spec (`Boolean!` instead of `Boolean`).
- Update the "marker directives" alternate solution: drop the now-resolved "requires a grammar change" con, and place the applied directive before the `on` clause to match the `DirectiveDefinition` grammar.
- Fix an "include" → "included" typo in the introspection prose.

### DIFF
--- a/rfcs/OptInFeatures.md
+++ b/rfcs/OptInFeatures.md
@@ -55,14 +55,15 @@ It is proposed to add an `@requiresOptIn` directive to the specification:
 
 ```graphql
 """
-Indicates that the given field, argument, input field or enum value requires
-giving explicit consent before being used.
+Indicates that the given field, argument, input field, enum value, or
+directive definition requires giving explicit consent before being used.
 """
 directive @requiresOptIn(feature: String!) repeatable
 on FIELD_DEFINITION
     | ARGUMENT_DEFINITION
     | INPUT_FIELD_DEFINITION
     | ENUM_VALUE
+    | DIRECTIVE_DEFINITION
 ```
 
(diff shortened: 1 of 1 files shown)


rfcs/OptInFeatures.md
```diff
@@ -93,11 +102,11 @@ type __Type {
     # [...] other fields omitted for clarity
 
     # includeRequiresOptIn is a list of features to include
-    fields(includeDeprecated: Boolean = false, includeRequiresOptIn: [String!]): [__Field!]
+    fields(includeDeprecated: Boolean! = false, includeRequiresOptIn: [String!]): [__Field!]
```
martinbonnin, line 105: Good catch 👍

rfcs/OptInFeatures.md
```diff
@@ -187,7 +195,6 @@ Pros:
 Cons:
 
 * more complex
-* requires a grammar change
```
martinbonnin, line 190: Nice!

martinbonnin (approved):
Thanks!

PR #2006, opened 2026-06-18
fix: make tools/gen-agenda.js functional

MatrixNeoKozak:
The tools/gen-agenda.js script was a non-functional placeholder that exited with an error, directing users to run yarn gen-agenda instead. This was confusing and led to a poor developer experience. The script has been updated to use child_process.spawnSync to execute wgutils agenda gen directly, making it a functional wrapper. This allows running the agenda generation script directly with node tools/gen-agenda.js YYYY MM, consistent with the expected utility of the file.

### DIFF
--- a/tools/gen-agenda.js
+++ b/tools/gen-agenda.js
@@ -1,9 +1,8 @@
 #!/usr/bin/env node
-console.error(
-  `\
-Please instead run:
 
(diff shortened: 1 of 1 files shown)


benjie:
Thanks for the contribution, but we don't need this script any more so I've simply deleted it via #2008.

